PR #2, opened 2016-02-05
General cleanup

bonpixel:
Fixing a few things and adding a shrinkwrap since the new less version makes everything wonky. For now were are locking at 2.5.3 until we can sort it all out.


agats:
:+1:
Good idea with the shrink wrap. Also, out of curiosity did the `npm shrinkwrap` automatically alphabatize the json?

bonpixel:
:confused:  Huh, I didn't notice that until you pointed it out. I'm not sure but that seems like a good guess.

codetothepoint:
I remember having issues without shrinkwrap alphabetization in the past.  A lack of alphabetization caused severe merge conflicts.

codetothepoint:
But, ya - go for it!  Merge away!

robatron:
FEBS has a shrinkwrap stabilizer, `febs-shrinkwrap`, to address that issue.

codetothepoint:
What does this Outlander know of FEBS....?  ;-)

PR #6, opened 2016-02-06
Removing Docs from mainline

bonpixel:
This branch removes the docs completely from the mainline branch. The
rationale behind this is we are leveraging GitHub pages for the
documentation of this project. Instead of having documentation in two
places and a convoluted documentation system, we are opting to house the
docs in an orphan branch -per the recommended procedure on GitHub https://help.github.com/articles/creating-project-pages-manually/- and
all documentation efforts need to be edited then PR approved to be
merged into the `gh-pages` branch. Indeed this will require a little
different workflow but in the end, I think this will greatly simplify
things and it should also prevent wasted effort. An added bonus is it
keeps the mainline a little cleaner which is kinda nice. Questions and
comments welcome.


bonpixel:
@mhewson @agats Take a peeksie if you could.

bonpixel:
hmmm this may be totes dumb
stand by

robatron:
The way I've done this before, I've kept the Jekyll source of the docs in `master` like it is now, made the `gh-pages` branch mirror `master`, and configured Jekyll to look in the docs directory. That way, you don't have to maintain a split codebase across two branches. (Which is sort of an antipattern of branching anyway.)

bonpixel:
Yeah we were thinking the same thing on Friday. I am going to close this

PR #7, opened 2016-02-08
Large amounts of whitespace in code blocks

mhewson:
The current version of rouge/liquid seems to add more whitespace to code blocks that is waned - 
[issue 216](https://github.com/Shopify/liquid/issues/216), for now reformatting the code samples to eliminate as much as possible


codetothepoint:
These changes look good.  It looks like our friends at Shopify are working to correct this.  Merge away!

PR #8, opened 2016-02-08
Large amounts of whitespace in list-group code blocks

mhewson:
The current version of rouge/liquid seems to add more whitespace to code blocks that is waned - 
issue 216, for now reformatting the code samples to eliminate as much as possible

### DIFF
--- a/_includes/collections/list-groups/css-list-group.html
+++ b/_includes/collections/list-groups/css-list-group.html
@@ -6,9 +6,9 @@
   <div class="well" data-example-id="simple-list-group">
   {% include /markup-templates/list-group/list-group.html%}
   </div>
-  {% highlight html %}
-  {% include /markup-templates/list-group/list-group.html param="default"%}
-  {% endhighlight %}
+{% highlight html %}
+{% include /markup-templates/list-group/list-group.html param="default"%}
+{% endhighlight %}
 
   <h3 id="list-group-sort" class="status status-warning">Sorting example</h3>
   <p>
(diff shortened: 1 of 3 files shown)


agats:
:+1:

robatron:
I don't think it's a good idea to change source code formatting for the sake of generated code, if that's the intention of this PR. Source code is for humans, generated code is for machines.

mhewson:
Unfortunately, this impacts the display of documentation. The  
{% highlight html %}
{% endhighlight %}
code blocks on some examples, add so much whitespace above, below and to the left of the code snippet that the page looks broken.  

see http://rei.github.io/rei-cedar/modules/tabs/ as a good example

robatron:
Gotcha, that's too bad. Templating languages are frustrating sometimes. Maybe rewrite everything in FreeMarker? ;)

mhewson:
Ha! yes

codetothepoint:
I tend to agree with Rob here.  Minor clean up is one thing, but our source should be as readable as possible.  We may want to wait for the fix to come from Liquid.

But, ya!  If only Freemarker was the universal language of templating...

wjramos:
It's been an open issue for 2.5 years -- @codetothepoint do you think it's going to be something to wait for?

robatron:
This PR fixes the rendering issue, workaround or not. No need to wait.

wjramos:
+1

mhewson:
The problem is that this solution would have us alter each code highlight example, seems like we should be able to do something like [ remove the whitespaces after all pages have been compiled](http://stackoverflow.com/questions/20314234/how-can-i-keep-jekyll-from-adding-whitespace-in-highlight)

robatron:
:+1: 

The [Python script solution](http://stackoverflow.com/a/20532291) would be simple to write in Node, which you could call after the docs are generated.

PR #33, opened 2016-02-17
Adopting a synchronous strategy

bonpixel:
This adds in a plugin that will allow us to more easily define
synchronous tasks for the gulp file. Until the release of Gulp 4.0 this
seems like a reasonable solution to serve our requirements for
synchronous thingies.

This also removes the duplicity around the `docs` task which was kicking off the `js` and `css` jobs a second time. The new workflow will be that docs only copies the existing compiled assets (if they are there). If one wants to bring in the compiled Cedar deliverables into the docs then they would need to run either the master `gulp` task or `js` and `css` individually.


bonpixel:
:boom: It builds!

PR #36, opened 2016-02-18
Fixing the file injection sequence for qunit

bonpixel:
There was a race condition where files were not being injected in the
correct order for qunit tests. This change returns the stream to the
injection that gulp actually knows when the task is done and should
prevent the race condition from happening.


wjramos:
Nice work, Approved.

PR #45, opened 2016-03-02
Flex grid

mhewson:
Nice!
do you see adding additional classes to our grid with this pr to allow for alternate flex display?
or will we open a new one to account for the new flows?

such as

  .col-xl-first {
      order: -1;
  }
  .col-xl-last {
     order: 1;
  }

.row-xs-top {
   align-items: flex-start;
}

.row-xs-center {
    align-items: center;
}

ect...


wjramos:
The only issue I'm seeing so far is column-span elements containing other column-span elements need to be flex as well ( see footer nav for example )

I think that the column grid makes most sense of equal-height inline that fill available space to best match the current behavior (I don't think row alignment makes sense presently). I do think that break-point order-changing would be pretty useful though.

mhewson:
Yea I see the need for break-point order-changing to allow for better content choreography.

wjramos:
I think the scope of this PR should be limited to ensuring all presently available column classes (col-xx-n, col-xx-push-n, col-xx-pull-n, col-xx-offset-n, etc.) work as expected. 

We also don't need clearfix anymore.

wjramos:
@bonpixel @mhewson -- Could you guys try building and see if you see any problems?

@mhewson I think this is a great use-case for running with casper

mhewson:
:+1: ApPrOvE lets get this in and cut a new minor

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/45/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

PR #46, opened 2016-03-03
Include normalize.css as a package

wjramos:
because it is something that is not intended to be modified, and include it in gulp less compilation. 

Also add option to cssnano to strip comments from minified generated file.


wjramos:
@bonpixel We good?

wjramos:
I self-approve

PR #69, opened 2016-03-17
replace pull-* classes to be compatible with flex grid

wjramos:


mhewson:
Will this cause problems for people using these classes outside of the flex grid - for instance within a float group?
My thought was that the col-(size)-(number)-(pull/push) classes should be used on grid and pull-\* classes should be used within content (cards) 

any ideas?

PR #72, opened 2016-03-18
Remove bootstrap from /src

wjramos:
- Moves bootstrap from /src into /node_modules
- Removes some unused dependencies
- Symlink from /src to /docs_src less
- Add 0.10.1 to .travis because we say we support it in our package.json
- Move normalize.css to app dependencies, rather than dev deps


mhewson:
approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/72/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '5'
```
wjramos, line 3: Changes in this file should be removed

mhewson:
you will want to babysit this into mainline when the next tag is created as it will break the build

PR #74, opened 2016-03-21
Alt grid

mhewson:


src/less/definitions/layout/grid.less
```diff
@@ -58,6 +63,10 @@
     }
 }
 
+.row.row_flex [class^=col-],
+.row.row_flex [class*=col-] {
```
wjramos: we're not really using underscores elsewhere, is this preferred over `.row.flex`, `.row.row-flex`?

mhewson: I'm fine with row-flex for now, but do think we should be moving to BEM over time

wjramos: But this isn't really BEM, is it? 

I thought the standard was:
`.block__element--modifier`

`.grid__row--flex` for example

mhewson:
Yes, I believe you are correct

mhewson:
all good?

mhewson:
approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/74/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

PR #79, opened 2016-03-23
Normin lite

mhewson:


bonpixel:
Simple. I like it. I Approve.

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/79/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

wjramos:
Wait, so you reverted it?

mhewson:
Yes, to be true to the intent of the tag, which was to introduce flex. While we still have the branch that integrates normalize as a dependency ready to go I wanted to be sure we would not rather have full control of our reset file first.

PR #83, opened 2016-04-05
adding interstate condensed

rcarlis:


src/less/definitions/elements/type.less
```diff
@@ -53,6 +53,18 @@
     ~"url('@{font-path}interstateBoldCondensed.woff') format('woff')",
     ~"url('@{font-path}interstateBoldCondensed.ttf') format('truetype')",
     ~"url('@{font-path}interstateBoldCondensed.svg') format('svg')";
+    font-weight: normal;
+    font-style: normal;
+}
+
+@font-face {
+    font-family: "InterstateCondensed";
+    src: url("https://satchel.rei.com/media/font/interstateCondensed.eot");
```
wjramos, line 62: This is only relevant for REI, see above for variable font path

`~"url('@{font-path}interstateCondensed.svg') format('svg')";` etc

mhewson:
I don't believe this should be an add to the pattern at this time, I will follow-up with Julie Simon, but my understanding is that the entire type system is changing shortly. This could be search specific at this time if they are requiring it for product tile work

wjramos:
@mhewson I think that as long as the fonts that are here are still here, it's okay to add instead of freezing all reasonable, but unknown-future additions.

I think it can be added, and then removed when we move all font-faces to another source, but until that time there is no sense in leaving something that currently fits here in purgatory

mhewson:
This is not a global pattern. This is an ask for a specific implementation on a vertical project. To my understanding design is not asking for this update nor are they interested in adding it for teams to use in a global capacity. If the Search team would like it they can maintain it within their codebase and account for understanding the ramifications of interstate being removed from the global type system.   At this time I have no understanding as to how this font fits into the Type system we provide globally

wjramos:
Right, but the added font face is a variation of one that already exists -- I understand what you are saying though. I don't think fontfaces should live in here at all, but until they are stripped out to live in themes, I think it's appropriate to build on what we have

mhewson:
Yes, I understand that.My thinking is that we have 3 fonts supported at this time which provide a hierarchy for our content. This addition has not been thought out or communicated or documented within that system.  what would this map to in the future font update? how would people be using that variable in code? the semantic meaning is not understood.  Anderson and Interstate bold condensed are used for specific content, this update enables a one off design to eek some extra px's in on something that should be accounted for with interstate bold condensed. This seems fine for a one off but when we make the update to new fonts we will end up trying to understand why our "random element" are mapping to something outside of the new visual language. 

If we want this in the framework we need to understand how it works within the whole system. I would assume at that time some elements that were scoped to use one of the other fonts would move to this fourth type.

PR #86, opened 2016-04-12
Removed prefixing from source

wjramos:


mhewson:
approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/86/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

src/less/mixins/vendor-prefixes.less
```diff
@@ -1,8 +1,12 @@
-// Vendor Prefixes
+// Shorthand prefixed properties
+//
+// While vendor prefixes are generated by Autoprefixer in Cedar, these mixins will remain intact
+// for those wishing to have shorthand prefixed properties without the need for Autoprefixer
+// in a consumer of Cedar.
+//
+// NOTICE: These mixins are deprecated for use within Cedar
```
wjramos, line 7: I may want to re-add these but keep the note to state that prefix mixins are not to be used within Cedar itself.

I guess the question of whether to leave as-is so that consumers using our Less, but not necessarily using autoprefixer (should we assume if they are compiling Less, that they should be also using autoprefixer?)

Either way, these mixins should remain present, even if for only utilities to access unprefixed properties so that current consumers are not affected by removal

mhewson, line 7: I don't have a problem with the idea of removing the prefixes and documenting that the consumer needs to account for them, outlining our recommendation. It seems like these mixins need to be relocated to a different file though as they no longer make sense in the context of vendor prefix

wjramos, line 7: Maybe we should split version compatibility items into its own area -- it's not stuff that should be in the trunk, but should exist just to not produce breaking effects

PR #87, opened 2016-04-12
Refactor a bit to consolidate less modules

wjramos:


mhewson:
looking good I would like to run a visual regression prior to taking these into master however

wjramos:
Yup, planning on it --

It should reduce file size of source and distributable files without affecting actual computed styles

src/less/definitions/collections/forms.less
```diff
@@ -114,97 +116,107 @@ label {
...
     }
-
     &[disabled],
     fieldset[disabled] & {
         cursor: @cursor-disabled;
     }
-
     // Reset height for `textarea`s
-    textarea& {
+    textarea & {
```
wjramos: I'll need to confirm this is correct

src/less/definitions/collections/forms.less
```diff
@@ -114,97 +116,107 @@ label {
...
-////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-//
-// Validation messages
-//
-// extends alert
-//
-//----------------------------------------------------------------------------------------------------------------------
-.form-control {
-    .rei-form-control-focus();
+    .input-lg + .form-control-feedback {
```
wjramos: Looks like this can be simplified as well

PR #89, opened 2016-04-14
Change the implementation of ratio items to ensure expected cross-bro…

wjramos:
…wser appearance when present inside of a flex layout. In a flex layout, flex items inherit their padding-bottom property from the height of the container, unlike block, so when there is no explicit height set the items have no height gained from the ratio. Example of problem here: http://stackoverflow.com/questions/23717953/padding-bottom-top-in-flexbox-layout

### DIFF
--- a/src/less/definitions/utilities/utilities.less
+++ b/src/less/definitions/utilities/utilities.less
@@ -257,13 +257,13 @@
  *
  ********************************************************************/
 
- .utility-aspect-ratio {
-     position: relative;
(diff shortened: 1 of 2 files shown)


mhewson:
approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/89/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

PR #92, opened 2016-04-14
Initial card work

wjramos:
Currently only considers a flex grid -- shouldn't be too far off to apply what we have here to standard float grid as well. All I know is for masonry layouts we will need to kill the floats.

Structural example: http://codepen.io/wjramos/pen/wGJvXo

Combined with img-frame: http://codepen.io/wjramos/pen/ZWvYVL


src/less/definitions/layout/cards.less
```diff
@@ -0,0 +1,84 @@
...
+
+    &.rounded {
+        border-radius: 10px;
+    }
+}
+
+
+.row-flex {
+
+    &.fixed {
```
wjramos: this may not be compatible with float

src/less/mixins/grid.less
```diff
@@ -74,6 +74,8 @@
 .make-row-flex {
     display: flex;
     flex-flow: row wrap;
+    justify-content: space-between;
+    align-items: flex-start;
```
wjramos: top align items

wjramos:
Proposed structure:

```
.container
    .row ( .fixed/.tile/.masonry/.overflow )
        .col-*
            .card
                ( header )
                    *
                section ( .row [ .fixed/.tile/.masonry/.overflow ] )
                    ( .col-* )
                        *
                ( footer )
                    *
```

mhewson:
can you link to the code pen for this one?

src/less/definitions/layout/cards.less
```diff
@@ -0,0 +1,84 @@
+
+.card {
+    display: inline-flex;
+    flex-flow: column nowrap;
+    position: relative;
+    overflow: hidden;
+
+    /* Contents should contain padding, etc. */
```
mhewson: where would this padding come from? This needs to be as intuitive and flexible as we can make it. What I would like to see is a utility class for card padding that can allow for multiple preset values.
maybe something like this?

class="card"
data-content-type="default, tile, panel" 
data-content-padding="[default, value-range]" 
data-content-gutter="[default, value-range]"

wjramos: Can you show me an example following best practice for this declaration style?

wjramos:
See OP for some examples -- it's not 1:1, but close

wjramos:
~~TODO: Add opaque background to card -- need to define what that will look like.~~

~~TODO: Need set of content padding classes for variants of card sections for space between edge of card and content~~ ( Repurposing `.well*` classes for this use )

mhewson:
I would think that should follow the existing well pattern

wjramos:
How do you mean? It does -- doesn't it?

wjramos:
@mhewson @bonpixel @rcarlis @panganibanpj @peripateticus @codetothepoint

wjramos:
cards should have an opaque background and border-radius ( unless row is set to `.tile` )

src/less/definitions/layout/cards.less
```diff
@@ -0,0 +1,84 @@
...
+
+    &.tile {
+        padding: 0 20px;
+
+        &.masonry {
+            padding: 0;
+            min-width: 100%;
+            margin: 0 auto;
+
+            .card {
```
wjramos: should un-nest this from masonry

src/less/definitions/layout/cards.less
```diff
@@ -0,0 +1,84 @@
...
+
+
+.row-flex {
+
+    &.fixed {
+        align-items: stretch;
+    }
+
+    &.tile {
+        padding: 0 20px;
```
wjramos: This will likely conflict with isotope/masonry layouts, need to confirm then find an alternative method to add spacing

src/less/definitions/layout/cards.less
```diff
@@ -0,0 +1,41 @@
...
+                border: none;
+                box-shadow: 0 @card-hover-offset 3px -1px rgba( 0, 0, 0, 0.35 );
+                transform: translateY( -@card-hover-offset );
+                text-decoration: none;
+            }
+        }
+    }
+
+    &.rounded {
+        border-radius: 1rem;
```
wjramos: This should be default unless `.tile` is applied to containing row

wjramos:
card row

```
<ul class="row row-flex fixed">
    <li class="col-xs-6 col-sm-4 col-md-3">
        <a href="#" target="_blank" class="card">
            <section class="img-frame center ratio-9-16">
                <img src="IMG"/>
            </section>
            <section class="well">
                <p class="label-classification">Label</p>
                <h3>Title</h3>
                <p>Text</p>
            </section>
       </a>
    </li>
    <!-- ... -->
</ul>
```

wjramos:
@mhewson -- could you vet these in parallel?

PR #96, opened 2016-04-22
Add support for application of breakpoint-specific ratios

wjramos:
Adds support for `.ratio-xs-[ratio]`, `.ratio-sm-[ratio]`, `.ratio-md-[ratio]` and `.ratio-lg-[ratio]`


src/less/definitions/utilities/utilities.less
```diff
@@ -270,132 +270,180 @@
...
+ }
+ .hidden-sm {
+   @media (min-width: @screen-sm-min) and (max-width: @screen-sm-max) {
+     .responsive-invisibility();
+   }
+ }
+ .hidden-md {
+   @media (min-width: @screen-md-min) and (max-width: @screen-md-max) {
+     .responsive-invisibility();
+   }
```
mhewson: don't these exist already in responsive-utilities.less?

wjramos: don't remember what I was thinking here

wjramos:
it would be a good idea here to force `height: 0` to avoid unintended layouts on elements that _do_ have implicit height

PR #98, opened 2016-05-12
Clean up flex grid, ensure that remnants from float grid not applied

wjramos:


mhewson:
Only thing I question is rather than defining this in row can we move this into the make-row mixin so we don't block someone from using the creation classes?

not sure about the longterm vision for them at the moment though so not a big deal

mhewson:
approve 

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/98/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

wjramos:
I'd rather see someone use `&:extend( .row-flex );` than using a mixin for it. It'll likely create code fragmentation otherwise.

I wonder, though, whether we should _require_ it to be `.row.row-flex` or if we should support simply `.row-flex` which would then nullify `.row` if present -- I'd imagine it would make it easier to port over existing classes down the road, where `.row-flex` becomes `.row` (still with `.row-flex` as an alias to not break existing usage ) when it becomes the standard

PR #99, opened 2016-05-18
Fix pattern nav

mhewson:
extending the mixin rather than using the mixin was breaking with the current build structure


mhewson:
approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/99/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

PR #106, opened 2016-05-24
Consolidate vertical rhythm

wjramos:


mhewson:
looks like this is covered in  a separate pr

PR #108, opened 2016-05-24
Consolidate hidden

wjramos:


src/less/definitions/utilities/utilities.less
```diff
@@ -32,7 +32,32 @@
...
+  }
+  &-md {
+    @media (min-width: @screen-md-min) and (max-width: @screen-md-max) {
+      .responsive-invisibility();
+    }
+  }
+  &-lg {
+    @media (min-width: @screen-lg-min) {
+      .responsive-invisibility();
+    }
```
mhewson, line 55: what is the benefit of this? why not just pass the .responsive-invisibility(); mixin to the .hidden class?

wjramos, line 55: There were no changes to the behavior -- it was just consolidated into a single block, with the intent of raising visibility of questionable implementations, your comment indicates that it is working already

mhewson:
gotcha approve 

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/108/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

PR #191, opened 2016-11-30
Docs/flash copy

lancetharper:
Removed zeroclipboard because it uses flash and isn't accessible which is a known issue  (https://github.com/zeroclipboard/zeroclipboard/issues/80). Will possibly revisit/replace with a full JS solution in the future.


megafatsquirrel:
Approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/191/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

PR #195, opened 2016-12-16
Fix/bold weight

lancetharper:
Fixes the safari issue of not rendering bold if you have the font installed.


src/less/themes/default/elements/typography.variables
```diff
@@ -30,7 +30,7 @@
 @brand-font-primary-name:                          "HCo Sentinel";
 @brand-font:                                       @brand-font-primary-name, @font-family-sans-serif;
 
-@brand-font-secondary-name:                        "Roboto Condensed", sans-serif;
```
mhewson, line 33: I worry about this not having a fall back, I would like it to be "Roboto Condensed", "Helvetica nue", sans-serif
does that cause a problem?

lancetharper: It shouldn't, it's just a fall back. Should I add Arial too like how it is a few lines up (commented out)?

mhewson: yes sir 
Roboto condensed -> roboto -> helvetica -> arial -> sans-serif

mhewson:
approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/195/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

PR #200, opened 2017-01-20
Adding in component compilation into Cedar

bonpixel:
This PR will allow a consumer to add require('rei-cedar') to their
project. Typically this will be used in a browserify or webpack type
manner but regardless, this provides the facility to require the
"kitchen sink" of the components (e.g. all of the defined components) in the Cedar repo.


bonpixel:
@mhewson @codetothepoint I think this will do the trick.

codetothepoint:
I think this is a great start!  Obviously we will need to move your components up here, and consider building a collection of artifacts (different tool shelves).  I say we merge this in, and start integrating the artifacts into our projects internally.  I'll get that going.

codetothepoint (approved):

bonpixel:
Ok, I ended up needing to rethink the way everything was being "bundled". Browserify and riotify were my natural first choice since that combo works so well in my project. Here things are a little different because we dont want to include the riot runtime (which gets added in via browserify) and the final bundle needs to look differently so that our users can take advantage of the `require('rei-cedar')` method of sucking this all in. 

It took a lot longer than I thought it would because I had to learn and then fitz with the node streams used in the gulp tasks. 

All in all I think it worked out pretty well and we should be able to continue to cut tags like we normally would to make everything jive.

bonpixel:
I should mention that I tested this branch against my project and was able to successfully use `require('rei-cedar')` and then the components mounted once my app bootstrapped...Kinda cool to see it all work.

If you have a riot app you can also test this branch via something like this:
`npm install git+https://github.com/rei/rei-cedar.git#component-deliverables --save`
then
`require('rei-cedar')` somewhere in a main index file.

mhewson:
great! approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/200/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

lancetharper:
Very cool. approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/200/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

lancetharper (approved):

PR #201, opened 2017-02-03
Created video modal and button components

matthewkoehl:
PR to review includes:
Setting up Video Modal and Modal button components
Styling close button with overlay for mobile


dist/rei-cedar-components.js
```diff
@@ -1,5 +1,108 @@
...
+      'button:not([disabled])',
+      'iframe',
+      'object',
+      'embed',
+      '*[tabindex]',
+      '*[contenteditable]'
+    ].join( ',' );
+
+    var tabTrap = function ( elem, evt ) {
+      var $elem = elem instanceof $ ? elem : $( elem );
```
Neogavin: Since this is gonna be called every keypress within the modal, you might want to put this inside of the `if()` statement below, so you're not creating a jQuery object every keypress.

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,174 @@
...
+      'button:not([disabled])',
+      'iframe',
+      'object',
+      'embed',
+      '*[tabindex]',
+      '*[contenteditable]'
+    ].join( ',' );
+
+    var tabTrap = function ( elem, evt ) {
+      var $elem = elem instanceof $ ? elem : $( elem );
```
Neogavin: Okay, my earlier comment should've probably gone here (its what I get for not looking through the whole PR first!).

src/components/modal-button/modal-button.tag
```diff
@@ -0,0 +1,8 @@
+<modal-button>
```
mhewson: One thing I would like to work on before commit is the naming conventions. Having a prefix like cedar would help @lancetharper thoughts?

src/components/video-modal/video-modal.tag
matthewkoehl: @Neogavin moved if statement to line 133, where call the tabTrap

src/components/modal-button/modal-button.tag
```diff
@@ -0,0 +1,8 @@
...
+  <button type="button" class="btn btn-primary" id="{opts.hashtag}" if="{ opts.cedarbutton == null || opts.cedarbutton == 1 }">{opts.buttontext}</button>
```
bonpixel: This file could be made clearer if the conditionals were first on each line.

bonpixel: Also feel free to add comments so that it's easy to parse. Riot should strip all of those on the compile step.

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,172 @@
...
+    <button id="modalCloseBtn" class="close icon icon-rei-close img-circle" aria-label="Close" onclick="{ this.hideModal }"></button>
+    <div class="modal-video-content position center">
+      <div class="embed-responsive embed-responsive-16by9">
+        <yield />
+      </div>
+    </div>
+    <a href="#modalCloseBtn" class="sr-only sr-only-focusable">End dialog</a>
+  </div>
+
+  <style type="less" scoped>
```
bonpixel: I've not seen two `<style>` declarations used in a single tag file before. Is this because you want some things scoped and some things unscoped?

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,172 @@
...
+        &.fade {
+          &.in {
+            opacity: 1;
+          }
+        }
+      }
+    }
+  </style>
+
+  <script>
```
bonpixel: Here again, two `<script>` block declarations. I think it would be best to just make these one block.

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,172 @@
...
+      'button:not([disabled])',
+      'iframe',
+      'object',
+      'embed',
+      '*[tabindex]',
+      '*[contenteditable]'
+    ].join( ',' );
+
+    var tabTrap = function ( elem, evt ) {
+      var $elem = elem instanceof $ ? elem : $( elem );
```
bonpixel: I don't know if the check you wrote is cheaper, but if you call `$( element )` on an already jquerifyied object, jquery is smart enough to not double jquerify it. e.g.

```javascript
var $elem = $( elem );
```

will do the same thing as the ternary here.

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,172 @@
...
+
+      // get list of all children elements in given object
+      var o = $elem.find( '*' );
```
bonpixel: You shouldn't need to query everything. Instead you should be able to do something like this:

```javascript
var focusableItems = $( focusableElements ).filter( ':visible' );
```

That will be cheaper since there could be N number of children in a model.

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,172 @@
...
+    trigger.addEventListener("click", function(e) {
+      that.showModal();
+    });
+
+    this.on('mount', function() {
+      setupModal();
+    });
+
+    function setupModal() {
+      var modal = document.getElementById('videoModal');
```
bonpixel: Why not reference the component itself `this.root`?

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,172 @@
...
+      modal.addEventListener('keydown', function(e) {
+        if (e.keyCode === 9) {
+          tabTrap(modal, e);
+        }
+      });
+    }
+
+    this.showModal = function() {
+      var body = document.querySelector('body');
+      var modal = document.getElementById('videoModal');
```
bonpixel: These can be optimized using `this.root`. No need to query from the document level.

Can these be wired up in the `mount` event? If these never change then rather than making all these queries everything you show the modal maybe we can just get them up front?

src/components/video-modal/video-modal.tag
matthewkoehl: Correct using that for scoping purposes

src/components/video-modal/video-modal.tag
matthewkoehl: The find, gets all the children of the element for filtering which I believe is actually what we want so as to be able to match the focus-able defined which are visible.

src/components/video-modal/video-modal.tag
matthewkoehl: I did a bit of re-working of the layout of the js in general, after thinking about it further, please let me know if any additional thoughts!

mhewson:
@bonpixel @Neogavin any other thoughts?

Neogavin (approved):

docs_src/_includes/header.html
```diff
@@ -12,7 +12,9 @@
     {{ page.title }} &middot; REI Cedar Style Framework
   {% endif %}
 </title>
-<script src="https://cdn.jsdelivr.net/riot/2.5/riot.min.js"></script>
+
+<script src="/assets/js/vendor/riot.min.js"></script>
```
megafatsquirrel: Does this URL point to the latest version of Riot? Might want to lock it down to a set version.

dist/rei-cedar-components.js
```diff
@@ -1,5 +1,115 @@
...
+        setTimeout(function() {
+          modal.classList.remove('show');
+          body.classList.remove('modal-open');
+          body.style.paddingRight = '';
+          body.removeChild(modalBackdrop);
+        }, 200);
+        modalLaunchId.focus();
+      }
+
+      var that = this;
```
megafatsquirrel: Is this variable used?

dist/rei-cedar-components.js
```diff
@@ -1,5 +1,115 @@
...
+        var modalBackdrop = document.createElement('div');
+        modalBackdrop.id = 'modalBackdrop';
+
+        modalBackdrop.classList.add('modal-backdrop');
+        modalBackdrop.classList.add('opaque');
+        modalBackdrop.classList.add('fade');
+        modalBackdrop.tabindex = -1;
+        body.appendChild(modalBackdrop);
+        body.classList.add('modal-open');
+        body.style.paddingRight = '16px';
```
megafatsquirrel: Can this be in a class? Seems like it could get lost when maintaining the code.

dist/rei-cedar-components.js
```diff
@@ -1,5 +1,115 @@
...
+        modal.classList.add('show');
+        setTimeout(function() {
+          modal.classList.add('in');
+          modalBackdrop.classList.add('in');
+        }, 100);
+        modal.focus();
+      }
+
+      function jsFriendlyHideModal() {
```
megafatsquirrel: Can we refactor the function name to hideModal().

src/components/modal-button/modal-button.tag
```diff
@@ -0,0 +1,46 @@
+<!-- When calling the tag, specify the opts attribute and pass to it a value from the avail below to choose which button to display -->
+<!-- If no value is specified, the button with the null condition will display by default -->
+<modal-button>
```
megafatsquirrel: I would love to have all riot component tag names be prefixed with cedar -> <cedar-modal-button>

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,222 @@
+<video-modal>
```
megafatsquirrel: Add the cedar prefix to the component tag name.

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,222 @@
...
+      'button:not([disabled])',
+      'iframe',
+      'object',
+      'embed',
+      '*[tabindex]',
+      '*[contenteditable]'
+    ].join( ',' );
+
+    var tabTrap = function ( elem, evt ) {
+      var $elem = $( elem );
```
megafatsquirrel: Are we using jQuery or should it not have any jQuery?

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,222 @@
...
+        // forward tab
+        // if focused on the last item and user preses tab, go to the first focusable item
+        if ( focusedItemIndex >= lastFocusableItem ) {
+          focusableItems.get( 0 ).focus();
+          evt.preventDefault();
+        }
+      }
+    };
+
+    var modal;
```
megafatsquirrel: Can this var be moved into the mount event function?

src/components/video-modal/video-modal.tag
```diff
@@ -0,0 +1,222 @@
...
+    this.on('mount', function() {
+      modal = this.root.querySelector('#videoModal');
+      var closeModal = this.root.querySelector('#modalCloseBtn');
+
+      // Function to show modal
+      function showModal() {
+        var body = document.querySelector('body');
+        var modalBackdrop = document.createElement('div');
+        modalBackdrop.id = 'modalBackdrop';
+        // IE workaround, list out each classList item
```
megafatsquirrel: What is the IE workaround? Is this related to a bug in IE or unsupported features in IE. Also do you know the version of IE.

src/components/modal-button/modal-button.tag
bonpixel: I agree. Should we do full `cedar-` or `cdr-`?

lancetharper: We had talked about using `cdr-` in the CSS so that would be my vote.

src/components/video-modal/video-modal.tag
mhewson: We will look at factoring jQuery out at a later time~

src/components/video-modal/video-modal.tag
matthewkoehl: http://caniuse.com/#feat=classlist  does not support multiple parameters

docs_src/_includes/header.html
matthewkoehl: Updated

dist/rei-cedar-components.js
matthewkoehl: Updated

dist/rei-cedar-components.js
matthewkoehl: updated

dist/rei-cedar-components.js
matthewkoehl: Stray var

src/components/modal-button/modal-button.tag
matthewkoehl: updated

src/components/video-modal/video-modal.tag
matthewkoehl: updated

src/components/video-modal/video-modal.tag
matthewkoehl: We're going to work towards not having it but talking with @mhewson, not worrying about it at the moment

src/components/video-modal/video-modal.tag
matthewkoehl: Need it defined outside

docs_src/_includes/header.html
```diff
@@ -12,7 +12,9 @@
     {{ page.title }} &middot; REI Cedar Style Framework
   {% endif %}
 </title>
-<script src="https://cdn.jsdelivr.net/riot/2.5/riot.min.js"></script>
+
+<script src="https://cdn.jsdelivr.net/riot/2.6.7/riot.min.js"></script>
```
bonpixel: Thanks for updating this but .com is on the path to upgrade to Riot 3 (there is an active PR for it) we should be using a Riot 3 version in this project: 

`https://cdn.jsdelivr.net/riot/3.3.1/riot.min.js`

In this context we're just talking about the cedar documents but check on the actual npm packages in the `package.json` and make sure that cedar is using a Riot 3 version to compile the tags preferably 3.3.1.

megafatsquirrel:
approve

[Approved with PullApprove](https://pullapprove.com/rei/rei-cedar/pull-request/201/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=rei/rei-cedar)

mhewson (approved):

PR #203, opened 2017-02-24
First pass at adding in the input component from OP

bonpixel:
Check this out @mhewson


megafatsquirrel:
Ouch

lancetharper:
@megafatsquirrel  Got rolled into #245 but in vue

PR #210, opened 2017-03-10
Media aggregator

matthewkoehl:
This PR encompasses the initial creation and styling of the cdr-event-aggregator tag, as well as implementing an example of its use in the /layouts/tiles/ section of the pattern library.


src/less/themes/default/layouts/groups.variables
```diff
@@ -46,6 +46,9 @@
 @group-grid-auto-flex-ruleset: {
     flex-grow: 1;
 };
+@group-grid-no-flex-ruleset: {
+    flex-grow: 0;
```
lancetharper: Setting flex -grow/-shrink/-basis individually like this is a little dangerous because browsers have different default values assumed for the others when only one is set. Unless we're only overriding (and can guarantee that somehow), I would stay away from defining those on their own like this.

src/less/themes/default/layouts/groups.variables
```diff
@@ -14,7 +14,7 @@
 @group-display:                                                 flex;
 @group-display-child-container:                                 inline-flex;
 @group-display-grandchild-flex-ruleset: {
```
lancetharper: Do we want to have mixins in our 'variables' files?

src/components/cdr-event-aggregator/cdr-event-aggregator.tag
```diff
@@ -1,22 +1,46 @@
...
             </div>
         </div>
     </section>
+
+    <style type="less">
+        .event__date {
+            padding: 1.2rem;
+        }
+
+        .event__date--title {
```
lancetharper: This isn't a modifier of event__date, it should be its own element (event__title?)

src/less/themes/default/layouts/groups.variables
matthewkoehl: Seems to be a common use throughout this groups.variables file... will need a bit of a general overhaul to get all out

lancetharper: Right, this is more of a general question. @mhewson ^^

src/less/themes/default/layouts/groups.variables
matthewkoehl: We are defining flex-grow on columns, so was an override, we might need to update if want to avoid this

lancetharper (commented):
Did we ever figure out how to make the heading level configurable for the event title?

PR #228, opened 2017-04-20
Implement button components from design

matthewkoehl:


src/css/components/button.pcss
```diff
@@ -1,18 +1,89 @@
-.cdr-button {
+.btn-group * {
```
lancetharper: If we're actually adding these styles, they need to have the cdr- namespace prefix and probably be camelcase

src/App.vue
```diff
@@ -27,10 +27,32 @@
...
+    <cdr-card className="cdr-card">      
+      <cdr-button disabled data-backstop="disabled">Disabled</cdr-button>
+    </cdr-card>
+    <cdr-card className="cdr-card">
+      <cdr-button data-backstop="primary-button-large">Primary large</cdr-button>
+      <cdr-button data-backstop="primary-button-small" :modifier="['small']">Primary small</cdr-button>
+      <cdr-button  data-backstop="primary-button-extra-small" :modifier="['extraSmall']">Primary extra small</cdr-button>
+    </cdr-card>
+    <cdr-card className="cdr-card">
+      <cdr-button data-backstop="secondary-button-large" :modifier="['secondary']">Secondary large</cdr-button>
```
mhewson: can we make the data attribute match the BEM name of the elements?

src/components/buttons/Button.vue
```diff
@@ -1,32 +1,30 @@
 <template>
-  <button :class="className" type="button">
+  <button class="cdr-button" :class="modifierClass" type="button" @click="handleClick">
```
mhewson: The Type attribute should probably be configurable to allow for submit / reset but default to button if nothing is specified.

src/css/components/button.pcss
```diff
@@ -1,18 +1,95 @@
...
-
-  &[disabled] {
-    color: black;
-    background-color: lightgray;
-  }
-}
+.cdr-btnGroup * {
+    width: 100%;
+    display: block;
+    margin: 0 0 20px 0;
```
mhewson: lets abstract out all of the values in these files 
everything  should be variable so a theme can set what they like.
also lets get the values using rem rather than px

src/css/components/button.pcss
```diff
@@ -1,18 +1,95 @@
...
+  &:hover,
+  &:active,
+  &:focus {
+    background-color: var(--btn-bgcolor-hover);
+  }
+
+  &[disabled] {
+    color: var(--btn-color-disabled);
+    background-color: var(--btn-bgcolor-disabled);
+    border-width: 2px;
```
mhewson: border width is set in cdr-button correct? should not need to define it here

mhewson: really the only change to these should be the colors

src/css/components/button.pcss
```diff
@@ -1,18 +1,95 @@
...
+    border-style: solid;
+    border-color: var(--btn-border-disabled);
+  }
+
+  &--secondary {
+    background-color: var(--btn-bgcolor-secondary);
+    border-width: 2px;
```
mhewson: same

src/css/components/button.pcss
```diff
@@ -1,18 +1,95 @@
...
+    border-style: solid;
```
mhewson: same

src/css/components/button.pcss
```diff
@@ -1,18 +1,95 @@
...
+  &:first-child {
+    margin-left: 0;
+  }
+}
+  
+.cdr-button {
+  color: var(--btn-color);
+  background-color: var(--btn-bgcolor);
+  border-width: 2px;
+  border-style: solid;
```
mhewson: variables for these values

src/css/components/button.pcss
```diff
@@ -1,18 +1,95 @@
...
+  &--small {
+    font-size: var(--btn-font-size-small);
+    padding: var(--btn-padding-small);
+    line-height: 1.5;
+  }
+
+  &--extraSmall {
+    font-size: var(--btn-font-size-extra-small);
+    padding: var(--btn-padding-extra-small);
+    line-height: 1.5;
```
mhewson: variable

src/css/settings/variables.pcss
```diff
@@ -30,10 +30,31 @@
...
+  --btn-color-blackAlternate: var(--color-architectural-5);
+  --btn-color-disabled: var(--color-architectural-4);
+  --btn-font-size : 21px;
+  --btn-font-size-small : 18px;
+  --btn-font-size-extra-small : var(--font-paragraph-font-size);
+  --btn-font-weight: var(--font-headline-font-weight);
+  --btn-line-height: 1;
+  --btn-padding-large: 13px 32px;
+  --btn-padding-small: 6px 24px;
+  --btn-padding-extra-small: 3px 16px;
```
mhewson: like the alphabetization

src/css/components/button.pcss
```diff
@@ -1,18 +1,95 @@
...
+      background-color: var(--btn-bgcolor-blackAlternate-hover);
+    }
+  }
+
+  &--whiteAlternate {
+    color: var(--btn-color);
+    background-color: var(--btn-bgcolor-whiteAlternate);
+    border-width: 2px;
+    border-style: solid;
+    border-color: var(--btn-border-whiteAlternate);
```
mhewson: one think I would like to see is that we alphabetize the value pairs within a selector - I find that is reduces the chances of overriding  value pair definitions on long sector chains.

src/components/buttons/Button.vue
```diff
@@ -1,32 +1,42 @@
...
-    data() {
-      return {
-        className: 'cdr-button',
-      };
+    props: {
+      modifier: {
+        required: false,
+        default: () => [],
+      },
+      type: {
```
mhewson, line 23: Would something like this work?
<template -- 
:type="type"
<script --
type: {
        type: String,
        default: 'button'
      },

mhewson: also <template 
:disabled="disabled"
and then 
<script 
  disabled: Boolean

PR #230, opened 2017-05-04
Feature/lists

matthewkoehl:


src/App.vue
```diff
@@ -61,6 +61,30 @@
...
+    <cdr-list :modifier="['bulleted']">
+      <li>List bulleted</li>
+      <li>List bulleted</li>
+      <cdr-list :modifier="['bulleted']">
+        <li>List nested bulleted</li>
+        <li>List nested bulleted</li>
+      </cdr-list>
+    </cdr-list>
+
+    <cdr-list type="ol" :modifier="['numbered']">
```
mhewson: can we have a default per type? it will be nice to not have to add the modifier unless we want something other than the normal expectation like ul default is bullet, ol default is numbered?

src/css/components/lists.pcss
```diff
@@ -0,0 +1,39 @@
+.cdr-list {
+  list-style-type: var(--list-style-default);
+  margin-bottom: 0;
```
mhewson: do we need to set margin here?

src/css/components/lists.pcss
```diff
@@ -0,0 +1,39 @@
...
+  padding-left: 0;
```
mhewson: lets try to be sure any values are being set with a variable

src/css/components/lists.pcss
```diff
@@ -0,0 +1,39 @@
...
+    padding-left: var(--list-padding-left);
+  }
+
+  &--menu {
+
+    & > li {
+      @apply --appearance-menu-list;
+      @apply --decoration-border-bottom;
+      
+      &:hover {
```
mhewson: this should come from the decoration/glaze

src/css/settings/appearance.pcss
```diff
@@ -1,20 +1,20 @@
 /*
 *
-* REI Cedar apperance
-* applying apperance from the following variables
+* REI Cedar appearance
```
mhewson, line 3: :0

src/css/settings/decoration.pcss
```diff
@@ -25,6 +25,11 @@
   --decoration-border-off: {
     border: none;
   }
+  --decoration-border-bottom: {
+    border-bottom-width: 0.1rem;
+    border-bottom-style: solid;
+    border-bottom-color: var(--color-medium-20);
```
mhewson: color should be from glaze

src/css/settings/font.pcss
```diff
@@ -21,6 +21,7 @@
 
   /* Size */
   --font-size-body-copy: 1.6rem;
+  --font-size-navigation: 1.8rem;
```
mhewson: once we pull future-vue into this branch lets reset these to use the values defined in brandai.pcss

src/App.vue
```diff
@@ -52,6 +52,30 @@
...
 
+    <cdr-list>
+      <li>List unstyled</li>
+      <li>List unstyled</li>
+    </cdr-list>
+    
+    <cdr-list :modifier="['bulleted']">
+      <li>List bulleted</li>
+      <li>List bulleted</li>
+      <cdr-list :modifier="['bulleted']">
```
lancetharper: Looks like nested lists do need to be wrapped in a list item. The a11y tests fail here.

src/App.vue
matthewkoehl: Yeah totally, think we talked about that the default for all lists was going to be the 'Unstyled'... but we can for sure do that way instead if you like?

src/css/components/lists.pcss
```diff
@@ -1,38 +1,50 @@
...
-  }
+  padding-left: var(--list-padding-left);
+  list-style-type: var(--list-style-bulleted);
 
   & > li {
     padding-bottom: var(--list-padding-bottom);
+
+    & > ul,
+    & > ol {
+      padding-top: var(--list-padding-bottom);
```
mhewson: maybe not  a great variable name?

src/css/components/lists.pcss
```diff
@@ -1,38 +1,50 @@
...
+      padding-left: var(--list-padding-left);
+
+      & > li:last-of-type {
+        padding-bottom: 0;
```
mhewson: confused as the why the padding above is using a variable and this is using a value?

src/css/settings/decoration.pcss
matthewkoehl: hmm, realized this color probably shouldn't live in decoration... So pulling it out - just wondering do we really need a glaze for just a border color? Looking at the examples of glazes so far, they seem to be encapsulating larger color schemes...?

src/components/lists/lists.play.js
```diff
@@ -0,0 +1,22 @@
...
+  .add('(Default)', {
+    template: '<cdr-list><li>unordered list</li><li>unordered list</li></cdr-list>',
+    readme: 'The default for an unordered list is bulleted.',
+  })
+  .add('Ordered List', {
+    template: '<cdr-list type="ol"><li>ordered list</li><li>ordered list</li></cdr-list>',
+    readme: 'The default for an ordered list is numbered.',
+  })
+  .add('Unstyled List', {
+    template: '<cdr-list :modifier="[\'unstyled\']"><li>unstyled list</li><li>unstyled list</li></cdr-list>',
```
mhewson: if you use 
`<cdr-list :modifier="['unstyled']"><li>unstyled list</li><li>unstyled list</li></cdr-list>`,

src/css/components/lists.pcss
```diff
@@ -0,0 +1,52 @@
...
+    list-style-type: var(--list-style-bulleted);
+  }
+
+  &--numbered {
+    list-style-type: var(--list-style-numbered);
+  }
+
+  &--menu {
+    list-style-type: var(--list-style-unstyled);
+    padding-left: 0;
```
mhewson: can we use variables rather than values?

PR #233, opened 2017-05-12
Feature/svg sprite

lancetharper:
This is a basic icon implementation, as we get more into other uses for them, we may want to revisit and add more a11y options. 

I tried creating the component to inline the svg (which is supported by every browser) but this made the exported js too large and made it difficult to support other icon sets.

Now, it fetches icons from brand.ai and compiles them into a single sprite. This sprite is then used as an external reference for `svg` with `use` and xlink:href. The library 'svgxuse' is included to add support for ie (and other browsers that may not correctly load them).

I've also fixed some broken backstop references in buttons and heading then got carried away and made all examples into components to help localize our testing grounds with their backstop files.


mhewson:
approve

PR #236, opened 2017-05-17
Future vue

matthewkoehl:
A couple variable values in glaze weren't matching, quick update to get those synced and resolve compiling warnings.


lancetharper:
I don't think you mean to merge this into master

PR #246, opened 2017-06-15
Vue-styleguidist updated, and some examples added

rafaesc:


rafaesc:
I had to rename the Image file because a Vue warning comes out, the component name can't be called Image

PR #256, opened 2017-07-06
Feature/updates

lancetharper:


mhewson:
what prompted the move from "yarn" to "npm run"?

lancetharper:
Using `yarn` sometimes fails the version check that runs before dev-server. It happens on and off with the yarn updates so I figure it's safer to just recommend the way that will always work.

PR #293, opened 2017-08-15
Update backstop, try adding it to travis

lancetharper:


lancetharper:
Never got this to work properly. Backstop should be doing a major version change soon with support for headless chrome -- will revisit then.

PR #324, opened 2017-10-03
adding cedar-fontless and font-caching js

ghost:
This adds a version of cedar that doesn't have fonts, and a javascript file that can be pulled to download the fonts async and then cache them in local storage for subsequent page views


lancetharper:
Would this work outside of rei.com as is?

mhewson:
have you considered using FontFaceObserver to ensure you display the down level font until we are sure that the new fonts are available? 
couple of examples 
https://www.zachleat.com/web/comprehensive-webfonts/#fout-class
https://www.filamentgroup.com/lab/font-events.html
https://calendar.perfplanet.com/2016/loading-web-fonts-asynchronously/

ghost:
@lancetharper yes it would

I've looked into FontFaceObserver, it acts as a full proof way of knowing a font has loaded and provides a way to normalize font-display behavior across browsers (which the above solution does as well). Its significantly larger to the above solution and additionally requires a promise pollyfill.

This would make me lean towards moving that type of implementation to the bottom of the document whereas this has a small enough foot print and is performant enough to have the option of inlining in the head of the document if we want to limit the flash of unstyled font if that bothers anybody (first page view only).

If we are cool with a bit of a longer flash of unstyled font / invisible font on the first view of a page then I think FontFaceObserver is a robust solution. We can definitely do both and see which one is more performant.

I've researched extensively and have read the above articles already. This solution implements a FOUT class

ghost:
Things worth mentioning / things that are different:

* Font will be loaded async, so flash FOIT/FOUT inevitable on the first view. Inlining the rei-cedar-font-loader.js limits this as much as possible.

* For modern browsers WOFF only support is suitable, any compression gains from WOFF2 would be negated by including it in addition to WOFF

* This will not work without javascript

peripateticus:
I forked FontFaceObserver and refactored so that it does not require a Promise polyfill if you wanted to give it a whirl it's here:  https://github.com/rei/fontfaceobserver/tree/v3.0.0. The code itself is the same size, it would just not need the Promise polyfill loaded in the head.  Note, the API for the load method is updated since you now need to pass in the success/fail callbacks.

ghost:
I removed the custom font loading dependency from cedar and not providing an
opinion on which font loader should be used and only providing the built font
CSS files to be used by a loader.

ghost:
This would not work as-is if the user chose to load the cedar-fontless bundle instead of the regular cedar bundle. They would  have to load the font css file seperately (ideally async)

I think if this works out, I could put efforts in cedar2 to make sure that a recommended loader is available and/or provide documented usage patterns.

mhewson (approved):

PR #372, opened 2018-02-08
Fix/grid gutters

lancetharper:
This allows us to support gutters that are greater than the container padding. It also removes the space that appeared above the first row in a cdr-row. A margin-bottom is added to each row to maintain the same total space as before (which we may decide to remove in the future).


mhewson:
whats the status of this PR?

mhewson (approved):
lets get rid of the bottom margin and add that in the row example class rather than as a row default

PR #411, opened 2018-04-25
Feat/plib 1871  link cedar

kevinmpowell:
@mhewson I still was unable to get `git cz` to work properly so I used the `--no-verify` arg to bypass whatever it was doing. I did get all my tests to pass locally except the visual diffing. Since this component completely changed (including its name) there's not really a baseline to visual diff against.

Mainly putting this up because I'm ready to receive more direct and actionable feedback on the code.


src/components/link/examples/Links.vue
```diff
@@ -0,0 +1,94 @@
+<template>
```
lancetharper, line 1: Just fyi, you can add `<!-- eslint-disable max-len -->` at the top of the template to disable lint errors in the template. Then you don't have to have all of those returns in the lorem ipsum :)

lancetharper:
Everything looks good so far. 

Only thing I see missing (beside icon stuff) is to add the link tokens to the token repo and consume them back in (https://github.com/rei/rei-cedar-tokens).

src/components/link/CdrLink.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <component
+    :is="el"
```
mhewson: Whats the consensus on this one? In cdr-text the `is` uses the term "tag". I am not opposed to "el" though do you feel it may be a bit confusing with how vue uses "el" to represent the DOM element being manipulated? Regardless of direction we want to keep consistency on the term - thoughts? @lancetharper @darintrimillos @kevinmpowell

src/components/link/CdrLink.vue
```diff
@@ -0,0 +1,74 @@
...
+ * @author [REI Software Engineering](https://rei.github.io/rei-cedar/)
+ */
+export default {
+  name: 'CdrLink',
+  components: {
+    CdrIcon,
+  },
+  mixins: [modifier],
+  props: {
+    el: {
```
mhewson: The documentation site will pull out comments for the props description, where appropriate add a description with the prop to call out what it is/ how to use it

src/components/link/examples/Links.vue
kevinmpowell: Thank you!

src/components/link/styles/CdrLink.pcss
```diff
@@ -0,0 +1,66 @@
...
+    outline: $outline-default-style $outline-default-width $color-outline-focus;
+  }
+}
+
+/* Style variants
+  ========================================================================== */
+
+/* On Dark
+    ========== */
+.cdr-link--on-dark {
```
mhewson: I feel that this is a fake design class at the moment we have not solved for themes and I would rather make a pass at link holistically to solve for `on-dark` `on-light` and such. can we explore solutions that would provide a similar result to your past color implementations for background/foreground color pairings?

src/components/link/styles/CdrLink.pcss
```diff
@@ -0,0 +1,66 @@
...
+    :Modifiers
+      :Style variants
+        :On Dark
+        :Standalone
+========================================================================== */
+
+.cdr-link {
+  background-color: $color-transparent-background;
+  border: 0;
+  color: $color-link--on-light;
```
mhewson: The variable should be managed from a link file that will need to be created https://github.com/rei/rei-cedar-tokens/tree/master/tokens and imported in here - 
@lancetharper was going to update the README on that repo to ensue the steps are outlined - sorry we have just started to move over to THEO so the docs may need some reworking to ensure we capture all the steps

src/components/link/CdrLink.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <component
+    :is="el"
+    :class="modifierClass"
+    :target="target"
+    :rel="computedRel"
+    :href="el === 'a' ? href : null /* don't include the href attribute if not an <a> */"
```
mhewson: Another stylistic question do we want to place the logic in the param like this or would it be more clean to do something like
`:href="url"` and then add that rendering logic to the `url` prop?

src/components/link/CdrLink.vue
```diff
@@ -0,0 +1,74 @@
...
+  },
+  mixins: [modifier],
+  props: {
+    el: {
+      type: String,
+      default: 'a',
+    },
+    href: {
+      type: String,
+      default: '#',
```
mhewson, line 35: the `#` standalone is not an accessible solution for this element. It would be rad to send a error/ warning to the console for this that asks that the implementers use the button element for links that will not have valid URL's  

I almost want to explore defaulting this the a `button` until they explicitly request a `href`

mhewson (changes requested):

src/components/link/CdrLink.vue
kevinmpowell: If there's a precedent with another component using `tag` and Vue ascribes special meeting to `el`, then we should definitely go with `tag`

kevinmpowell:
@mhewson and @lancetharper I believe I've made all the changes we discussed to Link and Icon. Let me know what next steps are.

lancetharper (changes requested):
Did we ever get tokens set up for link in the token repo? https://github.com/rei/rei-cedar-tokens

kevinmpowell:
> Did we ever get tokens set up for link in the token repo? https://github.com/rei/rei-cedar-tokens

@lancetharper can you help me understand what should go in that separate tokens repo and what should be in this repo's `src/css/` directory?
I'm confused by the definition of global CSS variables in both locations.

lancetharper:
@kevinmpowell 

`src/css` has some clean up to do still, sorry about the confusion.

In a nutshell: 

The global `src/css` stuff is for resets and **shared** classes (most likely utility type stuff)

The separate tokens repo is for **component specific** things (and foundations that we will share out via a variables file). 

It's mostly just us devs working in the tokens repo currently, but we'd like to get designers working in there too in the future which is why we have it isolated from the main code base. That and if we ever get to distributing these out for consumption on other tech stacks we have an avenue to do so.

I've tried to get steps for working with the tokens captured here: https://paper.dropbox.com/doc/Design-tokens-T3NsqojMfc8q4KQj37STu

kevinmpowell:
@lancetharper Travis seems to be failing based on insufficient test coverage in `CdrInput` is that something I should try to fix as part of this PR?

PR #419, opened 2018-05-11
Feat/list

mhewson:
probably want to move the content code over to tokens~


lancetharper:
What's up with all of the old commits from cedar 1 in here?

Looks like master was merged into this branch...was that intentional?

lancetharper:
Looks like one of the unit tests for lists needs to be updated.

This breaks activity card too because it was using inline list

PR #425, opened 2018-05-23
feat(docs): to increase discoverability of the component test pages, …

zfolwick:
…a link to them should be present.

image


lancetharper (changes requested):
You should use router-link instead of hard coded `<a>` since we're using vue-router

https://router.vuejs.org/en/api/router-link.html

lancetharper (approved):

PR #428, opened 2018-05-24
Feat/breadcrumb

KrisKnabel:
The breadcrumb component is ready for review.  Design has taken a look at it and after feedback and refactoring, they have given it the thumbs up.  100% unit test coverage of the breadcrumb has been achieved with no accessibility errors.


package.json
```diff
@@ -9,6 +9,8 @@
   "scripts": {
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "dev": "node build/dev-server.js",
+    "build": "cross-env NODE_ENV=prod node build/build.js",
+    "build:demo": "cross-env NODE_ENV=dev node build/build-demo.js",
```
lancetharper: What does this build:demo stuff do?

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -0,0 +1,129 @@
...
+  <!-- disable lint errors on line length in template -->
+  <!-- eslint-disable max-len -->
+  <div
+    ref="container"
+    :class="[modifierClass]">
+    <div class="ruler">
+      <span ref="ruler">
+        <span
+          v-for="(item, index) in items"
+          :key="index">
```
lancetharper: Keys need to be unique on the whole page (I think). This should have something more unique.

Vue has an internal _uid prop but it isn't recommended to use it in user code. We should probably make a mixin with a method for creating unique ids/hashes and use that for stuff like this since we'll need it more and more for a11y stuff. Many of the other components could make use of this too.

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -0,0 +1,129 @@
...
+          <span
+            class="delimiter"
+            v-if="index < items.length - 1">
+          /</span>
+        </span>
+      </span>
+    </div>
+    <div class="breadcrumb-container">
+      <span
+        v-if="(truncate)"
```
lancetharper: Can this just be `truncate`?

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -0,0 +1,129 @@
...
+      >
+        <span class="breadcrumb-item ellipses"><a
```
lancetharper: We should use BEM for all class names. See note in the css file

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -0,0 +1,129 @@
...
+          <a :href="items[items.length - 2]">{{ items[items.length - 2].displayText }}</a>
+        <span class="delimiter">/</span></span><span class="breadcrumb-item">
+          <a :href="items[items.length - 1]">{{ items[items.length - 1].displayText }}</a>
+        </span>
+      </span>
+      <span
+        v-else
+        class="breadcrumb-item"
+        v-for="(item, index) in items"
+        :key="index">
```
lancetharper: Again, should be unique

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -0,0 +1,129 @@
...
+<script>
+/**
+ * Cedar 2 component for breadcrumb
+ *
+ * @version 0.0.1
+ * @author [REI Software Engineering](https://rei.github.io/rei-cedar/)
+ */
+
+import modifier from 'mixinsdir/modifier';
+import _ from 'lodash';
```
lancetharper: Try and only grab what you need from lodash so we aren't getting everything but only using one part.

`import debounce from 'lodash/debounce';`

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
```diff
@@ -0,0 +1,57 @@
...
+    :Base - CdrBreadcrumb
+      :Elements
+        :breadcrumb-item
+        :delimiter
+      :Modifiers
+        :sm
+========================================================================== */
+
+.cdr-breadcrumb {
+  a {
```
lancetharper: Avoid using element selectors unless you **absolutely** can't. Everything we control should have class names so we don't collide with user styles.

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
```diff
@@ -0,0 +1,57 @@
...
+    @include redwood-display-10;
+
+    color: $breadcrumb-text-color;
+  }
+
+  /* Hover
+  ========== */
+  .breadcrumb-item {
```
lancetharper: We use BEM as a class naming strategy to keep our specificity low and control exactly what we mean to. https://css-tricks.com/bem-101/

Take a look at rating for a pretty simple example https://github.com/rei/rei-cedar/blob/cedar2/src/components/rating/styles/CdrRating.pcss

lancetharper (dismissed):

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: I was pretty sure the :key value scope is just for the collection but I could be wrong.  I would be surprised.  Better to be safe though so I like your mixin idea.

package.json
KrisKnabel: It is for building the components to host for a quick demo or design review.  Not a true prod build, sort of a hybrid between prod and dev.  Now that I think of it, I should put this in a different branch.  I will go ahead and do that now.

src/components/breadcrumb/CdrBreadcrumb.vue
lancetharper: I'm not sure if it's across the whole page, but using index alone is like not having a key at all since that's how vue would infer it anyway. At minimum I'd think it would need to have some contextual id like `breadcrumb${index}`

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: Tried that and it was not working, I'll try again.

src/components/breadcrumb/CdrBreadcrumb.vue
mhewson: here is a link the the start of our style guide - 
https://paper.dropbox.com/doc/REI-Cedar-code-style-guide-oHMNpqXcu1nvFLUI2pMUO

package.json
lancetharper: Ok, how is it different than `npm run dev`? I guess I'm not clear what a hybrid between dev and prod would do

KrisKnabel: npm run dev does not produce a dist folder.  When you build for prod, you lose the source mapping and can't debug on the hosted site.  The build:demo builds the dist folder while still allowing the user to see readable files and can debug in the browser.

KrisKnabel: I'm going to rip this out of this PR and keep it locally until it becomes useful again, say if design wants to look at changes on the fly or something.

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: Truncate is a computed property.  Made the change in response to your comments about not modifying the props directly.  Did I do it wrong?

lancetharper: No, I just wondered if it needed the `()` around it. `truncate` vs `(truncate)`

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: So according to the docs and the vue.js forum, it is actually better and more performant to not provide a key in this instance.  If we had child view components, we want to define :key but since this is a simple v-for list, we should just use Vue.js default implementation.  I feel that we need to modify the eslint to allow for this.

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: Ah, my guess is the parenthesis are not essential.  They were likely and remaining artifact from the previous comparison statement that was there before the refactor.  I will remove them.

src/components/breadcrumb/__tests__/CdrBreadcrumb.spec.js
```diff
@@ -0,0 +1,104 @@
+import { shallow } from '@vue/test-utils';
+import CdrBreadcrumb from 'componentsdir/breadcrumb/CdrBreadcrumb';
+import debounce from 'lodash/debounce';
+
+//jest.unmock('lodash/debounce');
+debounce = jest.fn((fn) => fn);
```
KrisKnabel: The lodash import change broke the unit test loader.  Still working on a way to get this to run.

package.json
KrisKnabel: The demo build logic has been removed.

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: This has been modified to use the most performant Vue.js default key logic.

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: The parenthesis have been removed.

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: Modified all class names to follow BEM guidelines

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: Modified to use Vue.js default logic for key

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: Updated dependency to only grab debounce function from lodash.

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
KrisKnabel: Updated class names to follow BEM guidelines.

lancetharper:
I pushed a small change to flatten out the css a bit and fix that button background color.

Only thing left to do is refactor some of the markup to make it more a11y friendly (nav + ol) and add a couple aria tags.

Check [here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html) for an example.

KrisKnabel:
Pushed changes suggested by Lance.  Ready for "hopefully" final review pass.

KrisKnabel:
Updated ellipses touch area, removed unnecessary div, updated unit tests, and updated example styling.  Please review.

KrisKnabel:
Updated ellipses to be a button with icon.  Only icon we had was vertically oriented so rotated it 90 degrees.  I was instructed not to translate it and keep it vertically centered(ish), but it looks odd IMHO.

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+  <!-- disable lint errors on line length in template -->
+  <!-- eslint-disable max-len -->
+  <!-- eslint-disable vue/require-v-for-key -->
+
+  <nav
+    ref="container"
+    :class="modifierClass"
+    class="cdr-breadcrumb__container"
```
lancetharper: Doesn't look like this class actually does anything. Can probably just be removed since the base class is added here now too.

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
```diff
@@ -0,0 +1,100 @@
...
+    cursor: pointer;
+    display: inline-block;
+    padding: 6px 11px; /* extend clickable area */
+    margin: -6px -11px; /* preserve layout */
+  }
+
+  /* Ellipses Icon
+  ========== */
+  &__ellipses-icon {
+    width: 1.4rem;
```
lancetharper: Overriding the icon size like this is load order dependent. If icon's css is loaded after breadcrumb, this won't do anything. We should be using the icons size modifier too rather than doing it this way.

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
```diff
@@ -0,0 +1,100 @@
...
+    flex: none;
+    color: $breadcrumb-text-color;
+
+    &:last-child {
+      color: $breadcrumb-last-child-text-color;
+    }
+
+    /* Hover
+    ---------- */
+    &:hover .cdr-breadcrumb__link {
```
lancetharper: Can we get rid of this descendant selector since __link is set to inherit color anyway? __link needs to have `color: inherit` added for it's hover state to do this.

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
```diff
@@ -0,0 +1,100 @@
...
+    @include redwood-display-10;
+
+    color: $breadcrumb-delimiter-text-color;
+    padding-left: $space-quarter-x;
+    padding-right: $space-half-x;
+  }
+
+  /* Ellipses
+  ========== */
+  &__ellipses {
```
lancetharper: Looks like this should underline for hover/active/focus

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
lancetharper: active and focus states should almost always match hover too

lancetharper:
Looks like the ellipse button needs screen reader text too to pass the a11y tests

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: Removed

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
KrisKnabel: Removed descendant selector and updated active and focus states

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
KrisKnabel: Had to add border-right since the svg does not respond to text-decoration and the svg was being rotated.

src/components/breadcrumb/styles/CdrBreadcrumb.pcss
KrisKnabel: Made selector more specific.

KrisKnabel: The icon size modifier was not working.  Setting the width and height directly was the only way I could figure out how to get the icon to show up at all.

KrisKnabel:
On Friday 6/15, pushed up SEO Refactor changes which included: 
  - Added ellipsis svg to use instead of rotated moreFunctions icon.
  - Hidden list items during truncation are now rendered and hidden so bots can still see them.
  - Modified breadcrumb collection to match JSON-LD format for breadcrumb for easier consumption
  - Updated examples to make it clearer to not use the current page/steps in breadcrumbs

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -0,0 +1,167 @@
...
+  props: {
+    /**
+     * Required. List of source breadcrumb property objects
+     *
+     * {
+     *   displayText: Breadcrumb Display Text
+     *   url: location breadcrumb should navigate on click
+     * }
+     */
+    items: {
```
lancetharper, line 91: Can we get some validation for these props? `items` will probably need to check that the Array is of objects and that the objects have the required properties.

lancetharper: I think we're all good after this 👍

KrisKnabel: Validator function added for `items` prop.  The other props shouldn't need a validator function as the `required` and `type` properties should be sufficient.

KrisKnabel:
Done refactoring this breadcrumb component.  Ready for final review.

lancetharper:
Looks like this needs a new unit test still

KrisKnabel:
Added unit tests for prop validator.  Now at 100%  coverage.

lancetharper (approved):

KrisKnabel:
Merging... Finally!

KrisKnabel:
Accidentally clicked close instead of merge button.

PR #430, opened 2018-05-30
Feat/css modules

lancetharper:
Apologies for the HUGE pull request here, but these are the highlights:

- CSS modules for all components (those currently merged into cedar2 branch)
- Jest --> Mocha (and fix all tests)
- Update vue-loader to v15 (and fix webpack configs for this)
- Update a bunch of dependencies (and fix any issues this caused)
- Column and Row moved to their own folder/package "Grid" and css for it lives there now too instead of in core
- Restore hot module loading 

If you want to build this branch, you'll want to delete `node_modules` and re `npm i`

You should also do this after this is merged.


src/components/button/styles/CdrButton.pcss
```diff
@@ -34,7 +34,7 @@
   text-decoration: none;
   white-space: nowrap;
 
-  &__icon {
+  & > :global(&__icon) {
```
KrisKnabel, line 37: Can you explain why you used the global scope here?  Not a critique, but more of a learning question.

lancetharper: `:global` tells css-modules to skip that class and leave it as is. In this case it's because the user will add `.cdr-button__icon` to the icon in button. If we made it into a css module, the user would have to maintain that class like `cdr-button__icon_x-y-z`

src/components/grid/CdrRow.vue
```diff
@@ -1,20 +1,11 @@
 <template>
-  <div
-    v-if="type === 'normal'"
-    class="cdr-row"
-    :class="rowClasses"
+  <component
```
KrisKnabel, line 2: `<component>` element???  What is this and how come Row is using it but Column is not?

src/components/grid/examples/Grid.vue
```diff
@@ -1832,7 +1832,7 @@ export default {
 </script>
 
 <style>
```
KrisKnabel, line 1834: Should this be `<style module>` ?

src/components/grid/styles/Grid.vars.pcss
```diff
@@ -0,0 +1,6 @@
+$rowGutter-xs: calc(16px);
+$rowGutter-sm: calc(16px);
+$rowGutter-md: calc(32px);
+$rowGutter-lg: calc(32px);
```
KrisKnabel: Why `calc()` for a constant?

src/components/link/CdrLink.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <component
     :is="tag"
-    :class="modifierClass"
+    :class="[$style['poo'], modifierClass]"
```
KrisKnabel: `poo`  ???

src/compositions/caption/CdrCaption.vue
```diff
@@ -1,12 +1,13 @@
 <template>
-  <div class="cdr-caption cdr-inset--squish">
+  <div :class="[$style['cdr-caption'], 'cdr-inset--squish']">
```
KrisKnabel, line 2: Are you missing `$style[` before `'cdr-inset--squish'` ?

KrisKnabel (commented):
Nice Work Lance!

src/components/grid/examples/Grid.vue
lancetharper: These styles are just for the examples so it doesn't matter

src/components/grid/CdrRow.vue
lancetharper: `<component>` is just a generic that we usually use with `:is` for changing what element it should be (in this case either `div` or `ul`). Column could use it, but it was easier to just leave it as is with the existing logic there.

src/compositions/caption/CdrCaption.vue
lancetharper: `cdr-inset--squish` is a global utility and not specific to caption so it doesn't need to be a module of it.

src/components/link/CdrLink.vue
lancetharper: Left over from testing 🤣

src/components/grid/styles/Grid.vars.pcss
lancetharper: Good find, this probably _was_ an actual calc at some point before we had design guidance.

src/components/button/__tests__/CdrButton.spec.js
```diff
@@ -1,28 +1,27 @@
-import { shallow } from '@vue/test-utils';
+import { shallowMount } from '@vue/test-utils';
```
mhewson, line 1: mocha vs jest change?

lancetharper: Nope, `test-utils` update

src/components/image/CdrImg.vue
```diff
@@ -146,34 +145,32 @@ export default {
...
+      classObj[this.$style['cdr-media-frame__cover--cover']] = this.cover;
       return classObj;
     },
     cropClass() {
       const base = 'cdr-media-frame';
       const cropArr = this.crop ? this.crop.split(' ') : [];
+      let final = [];
 
-      let final = '';
+      final = final.concat(cropArr.map(mod => this.$style[`${base}--${mod}`]));
```
mhewson: oh lord

src/components/mediaObject/package.json
```diff
@@ -30,14 +30,14 @@
     "@rei/cdr-assets": "^0.1.0",
     "@rei/cdr-link": "^0.1.0",
     "@rei/cdr-icon": "^0.1.0",
-    "@rei/cdr-img": "^0.1.0",
+    "@rei/cdr-img": "^1.0.0-alpha.1",
```
mhewson, line 33: whats the reason for the major alpha on the image? will we be moving other feature compete and stable components to this pattern?

lancetharper: You'll have to ask yourself, you did it 😆

src/components/image/CdrImg.vue
lancetharper: I'll refactor this to use the updated mixin method

src/components/quote/styles/CdrQuote.pcss
```diff
@@ -16,12 +16,14 @@
     font-family: Sentinel, Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif;
```
mhewson, line 16: should this file get an update to use token values?

mhewson (approved):

src/components/mediaObject/package.json
mhewson: I thought I recognized that :/

KrisKnabel (approved):
Massive PR looks good.  Nice work Lance!

src/components/checkbox/CdrCheckbox.vue
```diff
@@ -1,12 +1,12 @@
 <template>
   <!-- eslint-disable max-len -->
-  <div class="cdr-checkbox__wrap">
+  <div :class="$style['cdr-checkbox__wrap']">
     <label
       :class="[modifierClass, labelClass]"
```
KrisKnabel, line 5: Will we ever need to bind the class with the modifierClass?  e.g. `$style[modifierClass]`

darintrimillos (approved):

PR #436, opened 2018-06-15
Refactor/button

darintrimillos:
Icon-only button has been merged back into CdrButton.


src/components/button/CdrButton.vue
```diff
@@ -1,12 +1,14 @@
+<!-- eslint-disable max-len -->
```
mhewson: any reason we can not just wrap the offending text manually?

src/components/button/CdrButton.vue
```diff
@@ -1,12 +1,14 @@
...
 <template>
   <component
     :is="tag"
-    :class="[modifierClass, sizeClass, fullWidthClass, responsiveClass]"
+    :class="[modifierClass, sizeClass, fullWidthClass, responsiveClass, hasIconClass, iconOnlyClass, onDarkClass]"
```
mhewson: I understand hasIconClass and iconOnlyClass but I don't like that they are not one thing seems like too much focus on icons

src/components/button/CdrButton.vue
darintrimillos: No there isn't. Fixed.

src/components/button/CdrButton.vue
darintrimillos: Fixed - wrapped into one iconClass computed value.

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -14,23 +14,26 @@
         v-if="truncate"
         @click="shouldTruncate = false"
         :class="$style['cdr-breadcrumb__item']">
-        <cdr-icon-only-button
+        <cdr-button
```
darintrimillos, line 17: @KrisKnabel Updated breadcrumb to use cdr-button instead of cdr-icon-only-button which will be deprecated.

mhewson (approved):

mhewson:
https://docs.google.com/document/d/1e-iJVarPCUm7ytzWwloUKH4nQ_k-VZY88Ov7YiEy_6M/edit
you will want to version your component as alpha and work with test to validate this component on the cross browser tests

PR #454, opened 2018-07-13
Feat/tabs

KrisKnabel:
Ready for PR review(mostly).  Will push updates to unit tests soon.


lancetharper:
Visually this looks amazing! Nice work! 👍  I'll dig into it deeper today -- in the meantime it looks like you're missing a few lines of code coverage for unit tests.

src/components/tabs/CdrTab.vue
```diff
@@ -0,0 +1,88 @@
...
+  <!-- disable lint errors on line length in template -->
+  <!-- eslint-disable max-len -->
+  <transition
+    name="fly"
+    v-on="animationHooks">
+    <div
+      v-if="active"
+      :aria-hidden="!active"
+      :class="[modifierClass]"
+      tabindex="2"
```
lancetharper: I'm not sure why this is needed but changing tabindex > 1 is generally a bad idea. @mhewson thoughts?

src/components/tabs/CdrTabs.vue
```diff
@@ -0,0 +1,182 @@
...
+  },
+  mounted() {
+    if (this.tabs.length > 0) this.tabs[0].setActive(true);
+    // Check for header overflow on window resize for gradient behavior.
+    window.addEventListener('resize', debounce(() => {
+      this.headerWidth = this.getHeaderWidth();
+      this.calculateOverflow();
+    }, 250));
+    // Listen for left and right arrow keypress
+    window.addEventListener('keydown', debounce((event) => {
```
lancetharper: Looks like this makes all of the tabs on the page go left/right instead of just a single one, not sure if this is intended

lancetharper: Vue also has events for key presses if we want to try those -- https://vuejs.org/v2/guide/events.html#Key-Modifiers

src/components/tabs/CdrTabs.vue
```diff
@@ -0,0 +1,182 @@
...
+    }, 250));
+    this.headerWidth = this.getHeaderWidth();
+  },
+  updated() {
+    this.initializeOffsets();
+  },
+  methods: {
+    handleClick(tabClicked, event) {
+      const newSelectedTab = this.tabs.find(tab => tabClicked.name === tab.name);
+      this.tabs.forEach((tab, index) => {
```
lancetharper, line 94: Just throwing this out as a possible performance enhancement (I don't know if it would work here I just try to be wary of forEach if it isn't actually needed), but could we use `.some` or `.every` instead of `forEach` or do we really need to loop through all of them every time?

Here's a reference for the differences - https://stackoverflow.com/questions/7340893/what-is-the-difference-between-map-every-and-foreach

package.json
```diff
@@ -171,5 +171,8 @@
     ],
     "instrument": false,
     "sourceMap": false
+  },
+  "dependencies": {
+    "@rei/cdr-icon": "^0.2.1"
```
lancetharper: Looks like this was added by accident...?

lancetharper (commented):
I'd really like this to have some more a11y scrutiny from @mhewson since this is probably a complicated one. I think there might be some aria roles missing possibly (tablist, tabpanel, some aria-controls?). Here are the deque docs for tab panel -- https://www.deque.com/blog/a11y-support-series-part-1-aria-tab-panel-accessibility/

The other thing I noticed is if I click on a tab and the URL updates and I refresh I would expect the page to load with that tab selected but I just get a blank screen.

src/components/tabs/CdrTab.vue
KrisKnabel: Oh good catch.  Design wants the tab switching to occur with the left and right arrow keys and the tab functionality to ONLY tab between the active tab header item and the active panel instead of the default behavior where the tab button moves selection over the list of tab header items and the enter key activates it.  This is an artifact my attempts at moving the selection behavior to the active panel and skipping the natural tab headers that are next in line.  
TL;DR I will remove this!

src/components/tabs/CdrTabs.vue
KrisKnabel: I believe the intention is to never have more than one tabs component on a page.  BUT, there could very well be other components that listen to the left and right keypresses.  I could add logic to only handle the arrow keypresses if the focus is within the particular tabs component with maybe a caveat in the documentation that if a component needs to respond to the left right events that it MUST stop the event propagation.

package.json
KrisKnabel: Yep, I will remove this.

src/components/tabs/CdrTabs.vue
KrisKnabel: I can refactor this.  The iteration requires multiple truths to satisfy the completion of the logic.  I used .forEach because code readability is better and the performance gain with the more complicated solution would be negligible due to the number of tabs that would even fit on a page, but technically it is not the most efficient, so I will come up with something more efficient with good readability.

lancetharper: That's fine then, I wouldn't worry about it too much. I just try and keep my eyes open for stuff like this while I'm working.

package.json
KrisKnabel: Removed

src/components/tabs/CdrTab.vue
KrisKnabel: Changed tab index from 2 to 1

src/components/tabs/CdrTabs.vue
lancetharper: Design may recommend this -- but I would try and over build things as design often can't follow their own recommendations. Easier to fix now when no one is using it

zfolwick:
Tabs link and routes need to be added to the kitchen sink page: {domain}/tabs

PR #461, opened 2018-07-24
Cedar Data Table Component

allentayloriii:
Created a data table component for Cedar that supports different configurations and supports tables being created through feeding in data or creating their own mark up.


src/components/table/CdrTable.vue
```diff
@@ -0,0 +1,193 @@
+<template>
+  <!-- disable lint errors on line length in template -->
+  <!-- eslint-disable max-len -->
+  <div :class="$style['cdr-table__wrapper']">
+    <div :class="$style['cdr-table__scrollable']">
+      <table :class="[modifierClass]">
```
mhewson, line 6: should probably add in support for caption and summary

PR #469, opened 2018-08-01
accordion hide content fix

darintrimillos:
This fix hides the contents of a closed cdr-accordion-item so that the contents are not tabbable without breaking the animation.

fix(accordion): set visibility hidden on cdr-accordion-item slot contents when closed

affects: @rei/cdr-accordion


lancetharper (approved):

lancetharper:
Looks like you need a test for your event...?

PR #473, opened 2018-08-07
fix(breadcrumb): fixed console error where element was not available …

KrisKnabel:
Fix for defect PLIB-3659

affects: @rei/cdr-breadcrumb

added check to prevent failed attempt to retrieve offsetWidth of null

ISSUES CLOSED: PLIB-3659


mhewson:
We will need to get approval to merge in this update as this component has just been released.

PR #480, opened 2018-08-22
Feat/input

KrisKnabel:
Updated input to match new design spec here: https://airtable.com/tblvAr9Ne9NgyIC5T/viwHhAEyFN0TNFCcE

Removed validation logic and updated unit tests as well.

### DIFF
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -65,6 +65,7 @@
         :class="$style['cdr-input__post-icon']">
         <slot name="posticon"/>
       </span>
+      <small>This is helper Text</small>
     </div>
   </div>
 </template>
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -6,7 +6,6 @@
       :for="inputId"
       ref="label"
     >{{ label }}
-      <!-- @slot for required indicator -->
       <span
         v-if="required"
         :class="$style['cdr-input__required-label']">
@@ -203,5 +202,6 @@ export default {
 
 <style module>
   @import 'cssdir/settings/_index.pcss';
+  @import './styles/vars/CdrInput.vars.pcss';
   @import './styles/CdrInput.pcss';
 </style>
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -88,7 +88,7 @@ import propValidator from 'srcdir/utils/propValidator';
 export default {
   name: 'CdrInput',
   mixins: [modifier, size],
-  // inheritAttrs: false,
+  inheritAttrs: false,
   props: {
     /**
      * `id` for the input that is mapped to the label `for` attribute. If one is not provided, it will be auto generated.
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -128,11 +128,8 @@ export default {
     required: Boolean,
     /** @ignore */
     large: Boolean,
-  },
-  data() {
-    return {
-      focused: false,
-    };
+    /** @ignore */
+    value: String,
   },
   computed: {
     // Use given id or generate one
@@ -174,9 +171,6 @@ export default {
       this.$emit('input', value, e);
     },
     onBlur(e) {
-      this.$nextTick(() => {
-        this.focused = false;
-      });
       /**
       * Fires when input loses focus.
       * @event blur
@@ -189,7 +183,6 @@ export default {
       * Fires when input gains focus.
       * @event focus
       * @type {event} */
-      this.focused = true;
       this.$emit('focus', e);
     },
     onPaste(e) {
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -82,7 +82,7 @@ import propValidator from 'srcdir/utils/propValidator';
 /**
  * Cedar 2 component for input
  * **NOTE:** `v-model` is required.
- * @version 0.0.1
+ * @version 1.0.0
  * @author [REI Software Engineering](https://rei.github.io/rei-cedar/)
  */
 export default {
@@ -126,8 +126,6 @@ export default {
     disabled: Boolean,
     /** @ignore */
     required: Boolean,
-    /** @ignore */
-    large: Boolean,
     /** @ignore */
     value: String,
   },
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -5,87 +5,80 @@
       :class="labelClass"
       :for="inputId"
       ref="label"
-    >{{ label }}<span v-if="required">*</span>
+    >{{ label }}
+      <!-- @slot for required indicator -->
+      <span
+        v-if="required"
+        :class="$style['cdr-input__required-label']">
+        Required
+      </span>
     </label>
+    <!-- @slot for information -->
+    <span
+      v-if="$slots.info"
+      :class="$style['cdr-input__info-container']">
+      <slot name="info"/>
+    </span>
     <div :class="[inputWrapClass]">
       <textarea
-        v-if="multiLine"
+        v-if="rows && rows > 1"
+        :rows="[rows]"
         :class="[inputClass, modifierClass]"
         v-bind="$attrs"
+        v-model="value"
         :id="inputId"
-        :value="currentValue"
         @blur="onBlur"
         @input="onInput"
         @focus="onFocus"
         @paste="onPaste"
         @keydown="onKeydown"
-        :required="required"
         :disabled="disabled"
+        :required="required"
         :aria-label="hideLabel ? label : null"
-        :aria-describedby="errors.length ? messagesId : null"
         ref="input"
       />
-
       <input
         v-else
         :type="type"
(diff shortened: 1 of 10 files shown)


src/components/input/CdrInput.vue
```diff
@@ -5,11 +5,24 @@
       :class="labelClass"
       :for="inputId"
       ref="label"
-    >{{ label }}<span v-if="required">*</span>
+    >{{ label }}
+      <!-- @slot for required indicator -->
+      <span
+        v-if="$slots.required"
```
lancetharper: Is the required supposed to be configurable? If it only says "Required" you can just do that with css content and then this doesn't need to be a slot and required can just be a boolean.

src/components/input/CdrInput.vue
```diff
@@ -19,16 +32,14 @@
...
         :disabled="disabled"
         :aria-label="hideLabel ? label : null"
-        :aria-describedby="errors.length ? messagesId : null"
         ref="input"
       />
 
       <input
         v-else
-        :type="type"
+        type="text"
```
lancetharper: What about types other than text? email, password, search, etc.

src/components/input/package.json
```diff
@@ -33,5 +33,8 @@
   "devDependencies": {
     "@rei/cdr-assets": "^0.3.0",
     "pkg-ok": "^1.1.0"
+  },
+  "dependencies": {
+    "@rei/cdr-icon": "^0.2.1"
```
lancetharper: Should this be a peer & dev dependency or are we adding specific icons built in? Or should this be omitted entirely? I only see icons added in the slot which would make this unneeded.

lancetharper (commented):
Are the pre/post icons ever going to be clickable? I know they are in current designs. It might be worth following up with design to confirm because these may need to be buttons/anchors as well.

src/components/input/CdrInput.vue
KrisKnabel: There was a comment saying that we were to be only using text.  Setting type to email, will invoke the native browser validation.  I think I can add a property that disables this native validation.  You are correct, email, password, and search are probably best encapsulated in this input component and should be defined for accessibility.  I will add this type prop back and have 'text' as the default with a disclaimer that a type of *checkbox, radio, submit, select, button... etc* are not acceptable.

src/components/input/package.json
KrisKnabel: This is an artifact of a road I was previously going down and forgot to clean up.  Good Catch!  Will remove this.

src/components/input/CdrInput.vue
KrisKnabel: It wasn't specifically defined.  Didn't want to restrict it to be text.  There was mention that a popover could exist there as well, so using the slot was the best way I could think of to stay generic here.  I agree that this would be better as a boolean and design should be more decisive as to how it is indicating that an input is required.  I will consult Brian/Mary and see if I can persuade them on this.

KrisKnabel: Brian gave me the thumbs up to restrict this to the 'Required' text.  So I will switch this to a boolean.

src/components/input/CdrInput.vue
```diff
@@ -99,6 +88,19 @@ export default {
...
+     *  See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input for more details.
+    */
+    type: {
+      type: [String],
+      default: 'text',
+      validator: value => propValidator(
+        value,
+        ['text', 'email', 'number', 'password', 'search', 'url'],
+      ),
+    },
```
KrisKnabel, line 107: Listened with a screen reader and the input type definitely affects the screenreader output.  Type: text, email, number, password, search, and url seemed like the logical ones to go with.  Also added styling to hide the increment/decrement pseudo element on type="number"

src/components/input/CdrInput.vue
```diff
@@ -111,49 +113,15 @@ export default {
...
-      default: false,
-    },
+    rows: Number,
     /** @ignore */
     disabled: Boolean,
     /** @ignore */
     required: Boolean,
     /** @ignore */
-    immediateValidate: Boolean,
+    large: Boolean,
```
KrisKnabel: This will need to conform to whatever we come up with for size variants.

KrisKnabel: Changed to use size mixin/prop

src/components/input/CdrInput.vue
KrisKnabel: Listened with a screen reader and the input type definitely affects the screenreader output. Type: text, email, number, password, search, and url seemed like the logical ones to go with. Also added styling to hide the increment/decrement pseudo element on type="number"

KrisKnabel:
This branch contains changes from the patch/icon-inherit-color branch, so this branch/pr should be merged AFTER that branch

KrisKnabel:
Zach gave the thumbs up on the input component.  This is ready for a PR Review.

src/components/input/__tests__/CdrInput.spec.js
```diff
@@ -1,26 +1,22 @@
...
-  } else if (!isNaN(inputText) && inputText !== '') {
-    obj.state = 'error';
-    obj.message = 'Error Message';
-  } else {
-    obj.state = 'error';
-    obj.message = 'Error Message 2';
-  }
-  return obj;
-}
+// function validateFn(inputText) {
```
zfolwick: should this code be removed altogether?

KrisKnabel: Hmmm, that was removed.  Strange.  I will remove it and that below.

KrisKnabel: I just pushed with this removed, not sure why its not showing up

KrisKnabel: Looks like github just was taken some time.  The commented out code has been removed.

src/components/input/styles/CdrInput.pcss
```diff
@@ -29,26 +18,13 @@
...
 .cdr-input__label {
   @include maple-utility-20;
-  
-  color: $input-label-color-text;
-  margin-bottom: $input-label-margin-bottom;
 
-  &--error {
-    color: $input-label-error-color-text;
-  }
+  color: $heart-of-darkness;
```
lancetharper, line 22: `$text-color-primary--on-light` or whatever it is now? All of these text colors should map to different tokens like that...but I guess we'll just get to these once those are all updated.

src/components/input/styles/CdrInput.pcss
```diff
@@ -89,20 +72,8 @@
...
-
-  &--warn {
-    border-color: $input-warning-border-color;
-    border-bottom-right-radius: 0;
-    border-bottom-left-radius: 0;
-  }
-
   &--preicon {
-    padding-left: $input-preicon;
+    padding-left: calc($inset-half-x + $inset-quarter-x + 25px);
```
lancetharper: This is pretty specific 😄   is this how design described it?

src/components/input/CdrInput.vue
```diff
@@ -5,100 +5,107 @@
       :class="labelClass"
       :for="inputId"
       ref="label"
-    >{{ label }}<span v-if="required">*</span>
+    >{{ label }}
+      <!-- @slot for required indicator -->
```
lancetharper: slot?

src/components/input/styles/CdrInput.pcss
KrisKnabel: Yes, I figured that there was going to be a separate body of work.

src/components/input/CdrInput.vue
KrisKnabel: Leftover artifact that I missed from when we decided not to allow consumers to provide custom required markup.  Sorry, looked at this file so many times my brain glossed over it.  I will remove the comment.

src/components/input/styles/CdrInput.pcss
KrisKnabel: So this was due to having to `absolutely position(overlay)` the icons over the input because you can't actually insert icons inside an input.  So this pushes the input internal content over to make space for the icon overlay.  The `half-x` for the left side of the icon.  `25px` for the icon width, and `quarter-x` for the margin to the right of the icon.  `25px` I agree is not ideal, but should accommodate most icons.  I was going to call this out in the documentation.  Do you think there is a better way?

lancetharper: There probably isn't really a better way. Maybe just a comment explaining the hard coded value since we try to avoid those 👍

src/components/input/CdrInput.vue
KrisKnabel: Removed outdated comment.

src/components/input/styles/CdrInput.pcss
KrisKnabel: Created variable $input-pre-icon-default-width to store the input offset for when the $slot is used.

KrisKnabel:
Updated comments in reference to Lance's concerns about the pre icon width.  Ran tests and accessibility with axe and everything passed.  This is ready for review again.

lancetharper (approved):

PR #482, opened 2018-09-04
Refactor/cdr table

darintrimillos:
Still trying fix one test...


lancetharper:
Looks like the dependency checker is working 😃

PR #498, opened 2018-09-21
Feat/tabs

KrisKnabel:
Fixed tab panel content overflow defect that Zach discovered.

### DIFF
--- a/src/components/tabs/CdrTabPanel.vue
+++ b/src/components/tabs/CdrTabPanel.vue
@@ -82,6 +82,7 @@ export default {
       el.classList.remove(this.animationDirection);
     },
     handleUpArrowNav() {
+      console.log('KRISTEST this = ', this);
       this.$parent.setFocusToActiveTabHeader();
     },
   },
--- a/src/components/tabs/CdrTabPanel.vue
+++ b/src/components/tabs/CdrTabPanel.vue
@@ -8,8 +8,10 @@
       v-show="active"
       :aria-hidden="!active"
       :class="[modifierClass]"
-      tabindex="1"
+      ref="cdrTabPanelContainer"
+      tabindex="0"
       role="tabpanel"
+      @keydown.up.prevent="handleUpArrowNav"
       :key="name">
       <slot/>
     </div>
@@ -78,6 +80,9 @@ export default {
     setLeaveEnd(element) {
       const el = element;
       el.classList.remove(this.animationDirection);
+    },
+    handleUpArrowNav() {
+      this.$parent.setFocusToActiveTabHeader();
     },
   },
 };
--- a/src/components/tabs/CdrTabPanel.vue
+++ b/src/components/tabs/CdrTabPanel.vue
@@ -82,7 +82,6 @@ export default {
       el.classList.remove(this.animationDirection);
     },
     handleUpArrowNav() {
-      // console.log('KRISTEST this.$parent = ', this.$parent);
       this.$parent.setFocusToActiveTabHeader();
     },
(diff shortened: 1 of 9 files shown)


src/components/tabs/CdrTabs.vue
```diff
@@ -128,10 +128,6 @@ export default {
         const elements = Array.from(this.$refs.cdrTabsHeader.children);
         this.underlineWidth = elements[0].children[0].offsetWidth;
         this.widthInitialized = true;
-        // Set focus to default Tab header
-        this.$nextTick(() => {
-          this.$refs.cdrTabsHeader.children[this.activeTabIndex].children[0].focus();
-        });
```
KrisKnabel, line 134: This was found to be unnecessary, so I removed it.

src/components/tabs/examples/Tabs.vue
```diff
@@ -3,13 +3,26 @@
...
+
+            I have a feeling too. What? Well, I figured, what the hell. Alright, we're the pinheads. C'mon.
+
+            Yeah, well, I still don't understand what Dad was doing in the middle of the street. Right. Lou, gimme a milk, chocolate. Lorraine, my density has popped me to you. Great good, good, Lorraine, I had a feeling about you two. Right. Well, Marty, I want to thank you for all your good advise, I'll never forget it.</p>
```
KrisKnabel, line 32: This was added to show how a tab panel behaves with content that overflows.

src/components/tabs/CdrTabs.vue
```diff
@@ -21,10 +21,11 @@
           ref="cdrTabsHeader">
           <li
             v-for="tab in tabs"
+            role="tab"
+            :aria-selected="tab.active"
             :key="tab.id"
             :class="[ tab.active ? $style['cdr-tabs__header-item-active'] : '', $style['cdr-tabs__header-item']]">
             <a
-              role="tab"
               :tabindex="[ tab.active ? 1 : -1 ]"
```
mhewson: why is the tab-index management required here?

KrisKnabel: Because design wanted focus to move from the `tab header` to the `tab panel` when a `tab` keypress is performed, AND that `arrow right` and `arrow left` would cycle between `tab headers`.  Without this logic, `tab` keypresses will iterate through ALL the `tab headers` before moving focus to the `tab panel`.

KrisKnabel: I am glad you brought this up.  I think we should propose a design change to where, the `arrow down` keypress moves focus from the `tab headers` to the `tab panel` and remove the `tabindex` changes.

KrisKnabel: So, I spoke with Brian, and we have decided to alter the Tabs keyboard behavior to what is described in the previous comment.

KrisKnabel: Finally completed the down arrow focus change logic.  Was quite a bit more involved, especially with the unit tests due to the parent/child coupling.

src/components/tabs/CdrTabPanel.vue
```diff
@@ -8,8 +8,10 @@
       v-show="active"
       :aria-hidden="!active"
       :class="[modifierClass]"
-      tabindex="1"
+      ref="cdrTabPanelContainer"
+      tabindex="0"
```
KrisKnabel, line 12: tabindex set here to make the active tab container a focusable element as part of the design spec.

mhewson (approved):

PR #499, opened 2018-09-27
Set unit npm script to --watch, added notes to readme

darintrimillos:
Makes unit testing better.


lancetharper (dismissed):

README.md
```diff
@@ -68,6 +68,10 @@ Runs [karma](https://karma-runner.github.io/1.0/index.html) unit tests. We're us
 
 Runs [Nightwatch](http://nightwatchjs.org/) end-to-end tests.
 
+`npm run unit`
+
+Runs unit tests. Keeps mocha active to watch for changes and re-runs tests. Use `it.only()` to watch just a single test.
```
KrisKnabel, line 73: How do you specify which single test?  Param to it.only()?  I think some additional details/instructions will be beneficial

darintrimillos: I'll add a better example

KrisKnabel: Does the watch function only run the test that was changed?  If so, that is really cool.

KrisKnabel: If you don't have to specify and run from the root that is.

lancetharper (approved):

PR #506, opened 2018-10-15
Feat/color

mhewson:


src/mixins/buildClass.js
```diff
@@ -0,0 +1,40 @@
...
+ * @mixin
+ */
+export default {
+  methods: {
+    /**
+     *
+     * @param {String} prop -- the prop to build classes from
+     */
+    buildClass(prop) {
+      const base = this.baseClass;
```
KrisKnabel: I know that `modifier.js` did not handle it, but we should handle the case where this.baseClass is undefined/null

src/components/rating/styles/vars/CdrRating.vars.pcss
```diff
@@ -0,0 +1,7 @@
+// Default theme color family
+//Content colors should be centralized and switched out via the theme file.
```
mhewson, line 2: @lancetharper, Is this something theme could accomplish? maybe the older themes that still exist in button?

lancetharper: It can't work the way it used to since that was dependent on having a custom loader so you could have separate css files per theme. We could maybe try and replicate that, but it would take some work.

lancetharper (approved):

src/components/rating/styles/vars/CdrRating.vars.pcss
KrisKnabel: I feel we may be able to solve this with the $style directive other than trying to overwrite css from file sources.

mhewson: It would be great to get to a point were the variables in the component would not be specific to a theme -- At the least I think I will add these in a central location

PR #509, opened 2018-10-17
chore(test-utils): update test-utils and fix tests that broke

lancetharper:
affects: @rei/cdr-breadcrumb, @rei/cdr-button, @rei/cdr-input


lancetharper:
I wanted some of the newer fixes for pagination so I just fixed the few tests that broke with the update

darintrimillos (approved):

PR #513, opened 2018-10-26
Feat/size

KrisKnabel:
Created size mixin and integrated it into the button, ratings, and icon components.  Input and Select components will get updated in separate PRs in order to reduce the size.


lancetharper:
I think this normalization stuff will be an awesome QoL update for consumers.

Can we add responsive sizes to icon/ratings as part of this PR like button has while we're in here?

KrisKnabel:
> I think this normalization stuff will be an awesome QoL update for consumers.
> 
> Can we add responsive sizes to icon/ratings as part of this PR like button has while we're in here?

Good Idea, it should only require css updates.  I'll go ahead and add that.

lancetharper (dismissed):

KrisKnabel:
Fixed merge conflict with spacing mixin

lancetharper (approved):

PR #515, opened 2018-10-31
Spacing utility classes

darintrimillos:
This branch add spacing utility classes to rei-cedar. The class names match the language used by design for an easy design to dev translation and can be applied responsively like `.inline-2-x@lg`.

The spacing classes these are replacing are still included in `cdr-assets` for backwards compatibility, but they are no longer documented in the README.

### DIFF
--- a/src/css/README.md
+++ b/src/css/README.md
@@ -77,15 +77,16 @@ Intended to be used for adding margin to the right of an element.
 
 #### Variants
 
-| Class            | property     | value     |
-|------------------|--------------|-----------|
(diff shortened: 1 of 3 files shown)


KrisKnabel (commented):
I vote for keeping the `.cdr` in the class name as it is more explicit for consumers.  Getting @mhewson to provide feedback is probably a good idea.

mhewson:
many examples in the kitchen sink used these classes as well as well as some components using these in code - has this been run against visual regression to verify these changes?

PR #516, opened 2018-11-05
WIP(icon): adding inheritColor prop

KrisKnabel:
Added inheritColor prop to icon.  This used to be part of modifier, but since there was only one modifier value, made sense to move to a boolean prop.

### DIFF
--- a/src/components/icon/CdrIcon.vue
+++ b/src/components/icon/CdrIcon.vue
@@ -2,7 +2,7 @@
   <svg
     xmlns="http://www.w3.org/2000/svg"
     viewBox="0 0 24 24"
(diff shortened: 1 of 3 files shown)


lancetharper (dismissed):
Thanks Kris!

lancetharper:
This needs a feat() or a fix() commit though so that it gets picked up in the changelog. I think WIPs get ignored @KrisKnabel

KrisKnabel:
Ya, I expected there to be something else needed, then I was gonna do a fix commit.  I'll make a comment change or something.

lancetharper (approved):

PR #517, opened 2018-11-06
Patch/size validator fix

KrisKnabel:
Noticed there were console errors in next for the size prop validator.  Created a more robust prop validator function in the size mixin.  This branch contains the patch/icon-inherit-color changes, so this branch should be merged AFTER that branch.

### DIFF
--- a/src/mixins/size.js
+++ b/src/mixins/size.js
@@ -12,7 +12,9 @@
  */
 import BuildClass from './buildClass';
 
-const sizes = ['small', 'medium', 'large'];
+const ValidSizes = ['small', 'small@xs', 'small@sm', 'small@md', 'small@lg',
+  'medium', 'medium@xs', 'medium@sm', 'medium@md', 'medium@lg',
+  'large', 'large@xs', 'large@sm', 'large@md', 'large@lg'];
 
 export default {
   mixins: [BuildClass],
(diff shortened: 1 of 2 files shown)


lancetharper:
I fixed console errors on the reset branch I'm working on with the propValidator function we have in utils if you want to just cherrypick this into yours for now 

a80e232713bd8d9e1c39923177dff0903d959eef

KrisKnabel:
I don't think this is a fix that should be part of the validator mixin, but I could be wrong.  I'll look at what you did and see if that makes more sense.

KrisKnabel:
So Lance had a more elegant solution using the prop-validator utility he created.  I copied/cherrypicked his solution from 

https://github.com/rei/rei-cedar/commit/a80e232713bd8d9e1c39923177dff0903d959eef

KrisKnabel:
This is ready for PR Review

src/mixins/size.js
```diff
@@ -23,7 +22,10 @@ export default {
     size: {
       type: String,
       default: 'medium',
-      validator: value => (sizes.indexOf(value) >= 0) || false,
+      validator: value => propValidator(
+        value,
+        ['small', 'medium', 'large'],
```
zfolwick, line 27: @mhewson - didn't Nathan Curtis specifically warn us against using "small/medium/large"?

zfolwick: should there be an "extra-small"?

KrisKnabel: there is an extra-small(xs) breakpoint, but we have decided that there is only `small, medium, large ` for sizes.  If someone REALLY REALLY wanted to have an additional size such as `Extra Small` or `Extra Large`, they could use the `modifier` mixin for that.

lancetharper (approved):

PR #521, opened 2018-11-09
Resets cleanup

lancetharper:
I removed selectors that we weren't using from resets and relocated what I could into the components that relied on them.

There are no visually breaking changes (or any visual changes at all really) as a result of this for our components. I did my best to manually check these in all other supported browsers as well (though I would like to get another set of eyes on that to double check since manual testing can be error prone)

We may want to hold off on merging this until after the in flight stuff (data table, input, and tabs) are completed and I can pull those in to check them. I would prefer to merge this first then have those components double check themselves via their own before/after visual regression tests since I don't have a frame of reference for what "broken" would look like for those. Make sure to check in all browsers!

This does not include any defense for normalization css and recommend that as a next step before we actually publish the update to assets

### DIFF
--- a/src/components/tabs/CdrTabs.backstop.js
+++ b/src/components/tabs/CdrTabs.backstop.js
@@ -1,4 +1,6 @@
 module.exports = [{
   label: 'Tabs',
-  selectors: [],
+  selectors: [
+    '[data-backstop="tabs"]',
+  ],
 }];
--- a/src/css/generic/reset.pcss
+++ b/src/css/generic/reset.pcss
@@ -41,34 +41,6 @@ body {
   outline: none !important;
 }
 
-a {
-  color: $link-color;
-  text-decoration: none;
-  background-color: transparent;
-  -webkit-text-decoration-skip: objects;
-}
-
-a:hover,
-a:active {
-  color: $link-color;
-  text-decoration: underline;
-}
-
-a:not([href]):not([tabindex]) {
-  color: inherit;
-  text-decoration: none;
-}
-
-a:not([href]):not([tabindex]):focus,
-a:not([href]):not([tabindex]):hover {
-  color: inherit;
-  text-decoration: none;
-}
-
-a:not([href]):not([tabindex]):focus {
-  outline: 0;
-}
-
 a,
 area,
 button,
--- a/src/css/generic/reset.pcss
+++ b/src/css/generic/reset.pcss
@@ -41,82 +41,6 @@ body {
   outline: none !important;
 }
 
-hr {
-  -webkit-box-sizing: content-box;
-  box-sizing: content-box;
-  height: 0;
-  overflow: visible;
-}
-
-abbr[title],
-abbr[data-original-title] {
-  text-decoration: underline;
-  text-decoration: underline dotted;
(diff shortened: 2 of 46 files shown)


lancetharper:
I tried something different in this branch that may be wonky -- I worked in all WIP commits for the resets removal/relocation then squashed them into a single `feat` commit at the end so the message is kind of long. Not sure how this will affect the changelog, but we'll see.

KrisKnabel:
> I tried something different in this branch that may be wonky -- I worked in all WIP commits for the resets removal/relocation then squashed them into a single `feat` commit at the end so the message is kind of long. Not sure how this will affect the changelog, but we'll see.

When you say "squashed", are you referring to a `git rebase`?

src/components/tabs/CdrTabs.backstop.js
```diff
@@ -1,4 +1,6 @@
 module.exports = [{
   label: 'Tabs',
-  selectors: [],
+  selectors: [
+    '[data-backstop="tabs"]',
+  ],
```
KrisKnabel, line 5: Oops, my bad.  Thanks for adding that!

lancetharper:
> > I tried something different in this branch that may be wonky -- I worked in all WIP commits for the resets removal/relocation then squashed them into a single `feat` commit at the end so the message is kind of long. Not sure how this will affect the changelog, but we'll see.
> 
> When you say "squashed", are you referring to a `git rebase`?

Yeah, if you're doing it via cli that's how you'd do it. I just did it via the GUI I use

src/css/generic/reset.pcss
```diff
@@ -41,82 +41,6 @@ body {
...
-  bottom: -0.25em;
-}
-
-sup {
-  top: -0.5em;
-}
-
 a {
   color: $link-color;
   text-decoration: none;
```
KrisKnabel: `text-decoration: none` seems undesirable for a css reset on an anchor element.  I would think `underline` would be the default.  Am I missing something?

lancetharper: If we want it to match our default link style yes. But we often use anchors outside of just a simple text context that we probably don't want it underlined by default. I'm not sure how that would affect consumers though. Changing it breaks breadcrumbs, tabs, and linked ratings on our end -- which are easily fixable.

The hover and focus states are set to underline though which is also kind of weird now that I think about it. It should probably be the same in all states as a reset

I think I prefer no underlines because of the different ways they are often used, but I could really be swayed either way. Thoughts? @mhewson @darintrimillos

KrisKnabel: Ah I see.... makes sense.  Ya I could also go either way.

mhewson: I'm in agreement with Kris, my hope would be that we only deliver cedar expectations via the cdr namespace not through element selectors if possible. Could class .a or some crap like that could work for this if we dont want every one to use cdr-link?

lancetharper: Ok, I removed the anchor styles and fixed what broke.

KrisKnabel (approved):

mhewson (approved):

PR #532, opened 2018-11-30
Normalize and Cedar 1 defensive css

lancetharper:
Added defensive css against normalize and cedar 1 styles. Sometimes this meant increasing specificity, usually by using the same class (i.e. .cdr-button.cdr-button) but most of the time it was just adding it's own reset for a property.

Examples and their styles were updated to not be affected by normalize and cedar 1 css as well so there are no false negatives when something about formatting the layout of the component would fail.

Backstop tests pass both with and without cedar 1 styles and/or normalize.css included (components are now insulated from them)


src/components/grid/examples/Grid.vue
```diff
@@ -1957,5 +1957,14 @@ export default {
     text-align: center;
     padding: 10px;
     width: 100%;
+    line-height: 1;
+
+    & * {
+      line-height: 1;
+    }
+
+    & p {
```
mhewson: does this need to exist? I didn't think we were using generic elements in the examples any longer?

src/components/link/examples/Links.vue
```diff
@@ -131,6 +131,12 @@ export default {
 </script>
 
 <style>
+/** normalize line-height for consistent testing because links inherit it */
+
+.link-examples {
```
mhewson, line 136: shouldn't that come from the font inheritance in link

src/components/grid/examples/Grid.vue
lancetharper: Looks like we were in the grid examples, those can be removed though

mhewson (dismissed):

src/components/link/examples/Links.vue
lancetharper: It still does. If we don't have a reset around it then it will fail when testing vs cedar 1 styles because it will inherit from `body` instead and get false negatives. It would fail because it's working correctly 😄

mhewson: ahhhhh . snap

mhewson (approved):

KrisKnabel:
Noticed `line-height: 1`, `padding: 0`, and `margin: 0` quite a bit.  Moving forward on new components, should we be setting these to be on the safe side, or will we want to follow some other process for validating those needs?  Seems that reset/defensive css might be a new step/swim-lane in the development process that would likely happen close to accessibility validation.  Does that seem correct?  If so, is it as simple as doing regression testing against normalize and cedar 1?

KrisKnabel (approved):
Noticed line-height: 1, padding: 0, and margin: 0 quite a bit. Moving forward on new components, should we be setting these to be on the safe side, or will we want to follow some other process for validating those needs? Seems that reset/defensive css might be a new step/swim-lane in the development process that would likely happen close to accessibility validation. Does that seem correct? If so, is it as simple as doing regression testing against normalize and cedar 1?

lancetharper:
> Noticed `line-height: 1`, `padding: 0`, and `margin: 0` quite a bit. Moving forward on new components, should we be setting these to be on the safe side, or will we want to follow some other process for validating those needs? Seems that reset/defensive css might be a new step/swim-lane in the development process that would likely happen close to accessibility validation. Does that seem correct? If so, is it as simple as doing regression testing against normalize and cedar 1?

Yep. I've added both of those css files to the index.html and it's just a matter of uncommenting them and running backstop against it again.

PR #535, opened 2018-12-06
Component only css file

lancetharper:
`npm run component-css` to generate the css file which is output to `styles/component.css`

Maintains full compatibility with current component build.


KrisKnabel (approved):
Nice work!  👍

PR #537, opened 2018-12-06
feat(icon accordion): Update cdr-icon dependency

darintrimillos:
affects: @rei/cdr-accordion, @rei/cdr-link, @rei/cdr-media-object


KrisKnabel (approved):
👍

PR #556, opened 2019-01-30
Updates

lancetharper:
- Update to webpack 4 (and related loaders)
- Update vue test utils (and other testing related dependencies)
- Update linting (and fix new errors)
- Update postcss (and related plugins)
- Update lerna/lerna changelog
- Update babel
- remove lerna-semantic-release
- cleanup of unused dependencies


lerna.json
```diff
@@ -1,5 +1,4 @@
 {
-  "lerna": "2.11.0",
```
darintrimillos, line 2: I'm assuming if we don't specify a version it'll just use whatever is installed

darintrimillos (approved):
Updated and was able to build.

lerna.json
lancetharper: They don't call that out as an option in the docs so I just removed it. Doesn't seem to do anything

PR #563, opened 2019-02-13
feat(radius tokens): components using radius tokens have been updated…

mhewson:
… to use updated tokensv1 names

affects: @rei/cdr-button, @rei/cdr-card, @rei/cdr-checkbox, @rei/cdr-cta, @rei/cdr-img,
@rei/cdr-input, @rei/cdr-pagination, @rei/cdr-radio, @rei/cdr-select, @rei/cdr-activity-card

### DIFF
--- a/src/components/radio/styles/vars/CdrRadio.vars.pcss
+++ b/src/components/radio/styles/vars/CdrRadio.vars.pcss
@@ -1,6 +1,6 @@
 $radio-figure-checked-background-color: #3278ae;
 $radio-compact-figure-space-top: 2px;
-$radio-figure-radius: 50%;
+$radio-figure-radius: $cdr-radius-round;
 $radio-figure-hover-border-color: #616161;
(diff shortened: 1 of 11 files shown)


src/components/radio/styles/vars/CdrRadio.vars.pcss
```diff
@@ -1,6 +1,6 @@
 $radio-figure-checked-background-color: #3278ae;
 $radio-compact-figure-space-top: 2px;
-$radio-figure-radius: 50%;
+$radio-figure-radius: $cdr-radius-round;
```
KrisKnabel, line 3: Are we sure that `$cdr-radius-round` gets builts with a `%` as the unit?

KrisKnabel: Talked with you and you said it does in fact have the percentage as the unit.

KrisKnabel (approved):

PR #570, opened 2019-02-16
feat(motion): removing legacy motion token values, moving some values…

mhewson:
… to use token variables

affects: @rei/cdr-accordion, @rei/cdr-tabs

### DIFF
--- a/src/css/settings/options.pcss
+++ b/src/css/settings/options.pcss
@@ -379,22 +379,6 @@ $suede-shoes: #5197cd;
 $snap-decision: #e86868;
 
 /* ---------------------
-* Motion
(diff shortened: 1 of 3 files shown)


src/css/settings/options.pcss
```diff
@@ -378,22 +378,6 @@ $background-color-dark: #292929;
 $suede-shoes: #5197cd;
 $snap-decision: #e86868;
```
KrisKnabel, line 379: Do we still want `$snap-decision` and `$suede-shoes` available?

mhewson: we do until the designers account for them -- this branch is only for motion

PR #573, opened 2019-02-21
feat(tokens): add component token maps to cdr-assets

lancetharper:
affects: @rei/cdr-assets

This is the script I was talking about adding that looks for all *.vars.pcss files, copies them and creates an index file for easy consumption. I made the script separate in case we want to change this to it's own output outside of assets

### DIFF
--- a/src/cdr-assets/get-maps.js
+++ b/src/cdr-assets/get-maps.js
@@ -1,0 +1,20 @@
+const glob = require('glob');
+const fs = require('fs-extra');
+const path = require('path');
+
+module.exports = (outDir) => {
+  let index = '';
+  // get vars files
+  const files = glob.sync('../**/*.vars.pcss', { ignore: ['../**/node_modules/**'] });
+  // copy vars files
+  files.forEach((f) => {
+    const fname = path.basename(f);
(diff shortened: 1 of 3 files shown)


darintrimillos (approved):

src/cdr-assets/get-maps.js
```diff
@@ -0,0 +1,20 @@
...
+  // get vars files
+  const files = glob.sync('../**/*.vars.pcss', { ignore: ['../**/node_modules/**'] });
+  // copy vars files
+  files.forEach((f) => {
+    const fname = path.basename(f);
+    const ext = path.extname(f);
+    const outFname = fname.replace(ext, '.scss');
+    fs.copySync(f, `${outDir}/scss/${outFname}`);
+    index += `@import "./${outFname}";
+`;
```
KrisKnabel, line 16: Are line 15 and 16 correct?  Is it due to needing a newline in the concatenation?

mhewson (approved):

src/cdr-assets/get-maps.js
lancetharper: Yep, just for a new line

PR #590, opened 2019-03-07
chore(tabs): fixing github security vulnerability

KrisKnabel:
affects: @rei/cdr-tabs

removed the package-lock.json file for tabs as no other component tracks its own package-lock.json
file and github is saying there is a security vulnerability in the file.


KrisKnabel:
I decided to remove the file to be consistent since no other component has it's package-lock.json file checked in.  When I asked if we should be checking in the package-lock.json file, the team said yes, so I'm a little confused.  Alternatively I could update the file to fix github's security vulnerability complaint.

lancetharper:
That is weird that the others don't have a package-lock...but the correct resolution here is to upgrade lodash in package.json to 4.17.11

KrisKnabel:
Added the package-lock.json file back with the updated lodash version

lancetharper (approved):

PR #592, opened 2019-03-14
Feat/popover

darintrimillos:
Elizabeth has given a thumbs up, so here we go!


src/components/popover/CdrPopover.backstop.js
```diff
@@ -0,0 +1,4 @@
+module.exports = [{
+  label: 'Popover',
+  selectors: [],
```
lancetharper, line 3: No backstop tests? You can do clicks and stuff to launch the modal, but I don't remember exactly how to do that and we might need to wire that functionality up.

src/components/popover/CdrPopover.vue
```diff
@@ -0,0 +1,117 @@
...
+    hasDefaultStyling: {
+      type: Boolean,
+      required: false,
+      default: true,
+    },
+
+    /**
+     * Aria label - the popover title.
+     */
+    ariaLabelledBy: {
```
lancetharper, line 41: If these aria attributes aren't required we don't really need props for them (unless we need the defaults for basic a11y). If they are added by consumers they will just get appended to the root element anyway.

src/components/popover/CdrPopover.backstop.js
darintrimillos: I totally forgot about this. Will look into it.

mhewson: this is all for the pre v1 component correct?

PR #598, opened 2019-04-04
refactor(ui variables): first pass at button UI variables

cowills:
WIP, opening a PR to make it easier to leave comments


src/components/button/styles/CdrButton.pcss
```diff
@@ -23,53 +23,52 @@
...
+  padding: $cdr-button-primary-space-padding-lightmode;
+  text-align: $cdr-button-primary-text-align-lightmode;
+  text-decoration: $cdr-button-primary-text-decoration-lightmode;
+  text-transform: $cdr-button-primary-text-transform-lightmode;
 
   & > :global(&__icon) {
-    align-self: center;
-    fill: inherit;
-    margin-right: $cdr-space-quarter-x;
+    align-self: $cdr-button-primary-space-align-self-icon-lightmode;
```
mhewson: I think only the values that will change color will need the mode appended to the end. This is not a fully understood implementation though the idea is to set a parent class with a light or dark mode and have the styles available per element to account for that.

mhewson (dismissed):

cowills:
tested this with the kitchen sink and everything seems to be working as before.

left the theming for link TBD since it sounded like there were concerns about how to handle that.

mhewson (approved):

PR #601, opened 2019-04-10
[OLD] component variables

cowills:
This is an example of what I was suggesting we do wrt the UI variables and mixins.
I separated the "primary" styles from the "base" styles, and made it so that primary is still the default so nothing should change for the end user. I then copied those styles into mixins in the variables file.

So now a UI variable user could get a primary button by doing
`@include cdr-button-base-mixin; @include cdr-button-primary-mixin;`, while still having the option of using the UI variables directly.

i think this will make supporting UI variable consumers much easier, but it is a big change to the way we would write CSS for these components. LMK if you have any thoughts or concerns before i move forward with the tedious copy pasting 🤠

example sandbox demonstrating how 3rd parties can use these: https://codesandbox.io/s/jzv798low9


src/components/button/styles/CdrButton.pcss
```diff
@@ -23,110 +23,18 @@
...
-  cursor: $cdr-button-base-cursor;
-  display: $cdr-button-base-display;
-  outline: $cdr-button-base-outline;
-  overflow: $cdr-button-base-overflow;
-  padding: $cdr-button-base-padding;
-  text-align: $cdr-button-base-text-align;
-  text-decoration: $cdr-button-base-text-decoration;
-  text-transform: $cdr-button-base-text-transform;
-
-  & > :global(&__icon) {
```
cowills: these nested selectors will get a lil weird for UI variable consumers, so we will need to pull them out and handle them separately. Most of the nested selectors in button have to do with icons, so it should be easy to group that logic together

src/components/button/CdrButton.vue
```diff
@@ -2,6 +2,7 @@
   <component
     :is="tag"
     :class="[
+      primaryClass,
       modifierClass,
       buttonSizeClass,
```
cowills, line 7: can we de-dupe the size styling with mixins as well? currently each "size@breakpoint" class is defined separately, even though they are the same ("small@xs" and "small@sm" and "small@md" and "small@lg" all contain the same styles) However, the styling for the regular "small" class is different from the "small@breakpoint" styles.

src/components/button/styles/CdrButton.pcss
lancetharper: Icon is (necessarily) difficult in the mono repo. Anything to do with icon will most likely be much simpler once we're into the single package so some of this may go away.

src/components/button/CdrButton.vue
lancetharper: I believe @darintrimillos did most of the button work so he may be able to best speak to this.

src/components/button/styles/CdrButton.pcss
```diff
@@ -135,147 +47,102 @@
...
-
-    &:focus {
-      background: $cdr-button-icon-only-background-color-focus;
-      box-shadow: $cdr-button-icon-only-box-shadow-focus;
-      fill: $cdr-button-icon-only-fill-focus;
-    }
+    @include cdr-button-icon-only-mixin;
 
     &.cdr-button--on-dark {
       fill: $cdr-button-icon-only-fill-darkmode;
```
cowills, line 103: was not sure of the value of creating a "cdr-button-icon-only-darkmode-mixin" here for this 1 attribute ._.

PR #603, opened 2019-04-12
Assets/responsive

mhewson:


src/components/Utilities/demos/visibility.vue
```diff
@@ -14,16 +14,16 @@
       <thead>
         <tr>
           <th>class</th>
-          <th class="cdr-hide cdr-show@xs-only">
+          <th class="cdr-hide cdr-show@breakpoint--xs-only">
```
lancetharper: did we decide to go with `@breakpoint` 🤢

mhewson: Sure did - the thought was to be as explicit as possible

mhewson: anyone want to see something else?

src/components/accordion/examples/Accordion.vue
```diff
@@ -14,6 +14,7 @@
         Default
       </cdr-text>
       <cdr-accordion
+        spacing-class="cdr-space-mb-one-x@breakpoint--xs cdr-space-mb-four-x@breakpoint--sm"
```
lancetharper: What is the point of this `spacing-class` prop all over? Why wouldn't we just use `class`?

src/components/Utilities/demos/visibility.vue
lancetharper: I thought that the `@` made it pretty clear already

KrisKnabel: What are the other options besides '-only'? Is there also a '-greater' and '-less'?

KrisKnabel: Is there also a 'greater' and 'less'?

src/components/accordion/examples/Accordion.vue
mhewson: My thinking is that we want to have a easy to understand API for the components. If every component has the same basic set of pops (size, space...) then you don't really need to think of how to do it - 
For spacing specifically from what I was seeing some times we want the spacing class to be applied in a nested component structure / composition -- this would not be possible with the native class name attr.

KrisKnabel: What if we changed it from `spacing-class` to just `spacing`?  That would keep it consistent with `size`.  Alternatively we could just go with `space`.

src/components/Utilities/demos/visibility.vue
lancetharper: The breakpoints only go in one direction (up), `only` is just a convenience.

Want something large below `sm` but small above `md`? --> `large@xs small@md`

`large@xs` = large @ xs and up

which then gets changed `@md` by

`small@md` = small @ md and up

mhewson: yea we talked about this at the last dev meeting and that thought was that we have other use-cases for xs sm md lg. The solution proposed was to make breakpoint explicit so there we no confusion  cdr-btn-sm@md vs cdr-btn-sm@breakpoint--md

cowills: i think i was the one that brought up that "small@md" might get weird in variable names, but the variables arent exporting anything related to breakpoints it turns out >_<

i agree that "small@md" is better here

KrisKnabel: I think that this is unfortunately something that consuming devs will have mixed feelings about.  Some prefer verbosity and other don't.  IMO I feel most developers at REI would prefer clarity so I'm kinda leaning towards Hewson's changes.

mhewson: OK so you are all ok with the @sm ect?

KrisKnabel: Should we create a poll in slack and send it to consumers?

mhewson: :skull:

src/components/accordion/examples/Accordion.vue
cowills: presumably you could still just pass these directly into `class` if you preferred doing it that way, right?

i like the idea of having a way to group these classes logically, as it can be a bit weird using a dozen+ utility classes on something and having it just be one really long line of text under `class`. 

I dont have any opinions on the name other than it should be consistent (sizing/spacing, size/space, etc, etc,)

src/components/Utilities/demos/visibility.vue
cowills: isnt small@md what we're doing today? its kind of a breaking change to do it differently

src/components/Utilities/demos/inset.vue
```diff
@@ -1,84 +1,25 @@
 <template>
-  <div data-backstop="inset-space-utilities">
+  <div data-backstop="inline-space-utilities">
```
KrisKnabel: It appeared above that `inline` was removed, should this actually be `inset`?

mhewson: ah yea

src/components/icon/examples/Icons.vue
```diff
@@ -177,12 +177,6 @@ export default {
 <style>
   .icon-examples {
     line-height: 1;
-
-    .icon-hover {
-      &:hover {
-        fill: red;
-      }
-    }
   }
```
KrisKnabel, line 184: This seems unrelated to Assets.  Is this intended?

KrisKnabel (dismissed):

src/components/icon/examples/Icons.vue
mhewson: added that back in

KrisKnabel (approved):
👍

PR #606, opened 2019-04-26
refactor(tokens): updating the set values with tokens

mhewson:


mhewson:
not sure if this would be done the same way - but we are authoring the breakpoint doc https://docs.google.com/document/d/1d31wVHRbdb_HkEyGl6VdHorMnVXurcap5uxo3YrPCJ0/edit#
and I would like to have a working example of tokens being used in js~

KrisKnabel (approved):
👍

src/mixins/breakpoints.js
```diff
@@ -8,9 +8,11 @@
 /**
  * @mixin
  */
-const SMBreakpoint = 768;
-const MDBreakpoint = 992;
-const LGBreakpoint = 1200;
+import cdrTokens from '@rei/cdr-tokens/dist/js/cdr-tokens.common';
```
lancetharper, line 11: You should just be able to use '@rei/cdr-tokens` as that should be the default pointed to by main.

https://github.com/rei/rei-cedar-tokens/blob/114bc8fd1c289d442d0bf06412d35a79041c47b4/package.json#L19

cowills, line 11: should probably be importing the specific tokens rather than the whole object i.e, `import { CdrBreakpointSm, etc. } from '@rei/cdr-tokens/etc.'`. Might also need to import from the ES6 tokens build for it to tree shake out the un-used tokens.

PR #611, opened 2019-05-14
use v1 tokens in components and CSS reset

cowills:
Only re-mapped tokens which had a 1:1 value match in the new tokens that also matched the semantic intent.

Will work with designers to update or hardcode the questionable instances (marked TODO)


lancetharper (dismissed):

src/components/tabs/styles/CdrTabs.scss
```diff
@@ -131,6 +134,8 @@
     border: none;
 
     /* border: none; */
+    // TODO: this maps to cdr-color-text-link-lightmode but is being used for `background`
```
KrisKnabel: Yes, this just happened to be selected by design to use $easily-excited, but is not related to LINK, so NOT using cdr-color-text-link-lightmode is appropriate.

src/components/tabs/styles/CdrTabs.scss
```diff
@@ -131,6 +134,8 @@
...
+    // TODO: why is this `background` and not `background-color`?
     background: $easily-excited;
```
KrisKnabel: Your concern is valid and  `background-color` would be better here.  Just in case, I tested this locally and the underline behaves correctly with `background-color` instead of `background`.  👍 on changing it, or I can push up the change if you like.

cowills: just pushed up that change

KrisKnabel (approved):

PR #613, opened 2019-05-22
Refactor/component variables

cowills:
This PR should change absolutely nothing in the components except for how the CSS is assembled internally.

Tested on cedar-docs and kitchen sink to visually verify nothing changed in the components, as well as on climbers-site just to be sure.

published as `@rei/cedar@3.0.0-alpha-1` if you'd like to have a look yrself


mhewson:
@cowills whats the status of this branch?

src/components/button/CdrButton.vue
```diff
@@ -87,6 +88,9 @@ export default {
     baseClass() {
       return 'cdr-button';
     },
+    defaultClass() {
+      return this.modifyClassName(this.baseClass, 'primary');
+    },
```
KrisKnabel, line 93: Is `cdr-button` now always containing a `primary` modified class name?  It looks like `secondary` may just override this in the stylesheet?  Just double checking as it looks like primary is always in the classname.

cowills: yep, thats how it works, I believe that's how it was working in the css before as well.

we could refactor it in the future but i was trying to make as few changes to the CSS as possible

KrisKnabel: 👍

lancetharper (approved):

PR #615, opened 2019-05-23
Tab unit tests

lancetharper:
- Refactor tabs to work with SSR (remove `beforeMount`).
- Improve animation performance by using `transform: transistion` rather than `margin`. 
- Update unit tests for test-utils beta.29 and make use of more framework features/jest

### DIFF
--- a/src/components/tabs/CdrTabs.vue
+++ b/src/components/tabs/CdrTabs.vue
@@ -46,7 +46,10 @@
         />
       </nav>
     </div>
-    <div :class="$style['cdr-tabs__content-container']">
+    <div
+      :class="$style['cdr-tabs__content-container']"
+      ref="slotWrapper"
+    >
       <slot />
     </div>
   </div>
@@ -74,6 +77,7 @@ export default {
       underlineScrollX: 0,
       activeTabIndex: 0,
       widthInitialized: false,
+      headerWidth: 0,
       headerOverflow: false,
       overflowLeft: false,
       overflowRight: false,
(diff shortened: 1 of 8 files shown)


src/components/tabs/CdrTabs.vue
```diff
@@ -25,10 +25,10 @@
           ref="cdrTabsHeader"
         >
           <li
-            v-for="tab in tabs"
+            v-for="(tab, index) in tabs"
             role="tab"
             :aria-selected="tab.active"
-            :key="tab.id"
+            :key="tab.id ? tab.id : `${tab.name}-${index}`"
```
KrisKnabel, line 31: So I can learn, what is going on here?  It looks to me like a conditional binding where if `tab.id` is not null, than `tab.id` set to the value of `tab-name` + `-` + `index`?

lancetharper: Correct. It was trying to key by `id` but since `id` isn't a required prop it could have been empty if only `name` was supplied (which is required)

lancetharper: It's just an if statement -- if `tab.id` then use `tab.id` else use `tab.name...`

```condition ? exprIfTrue : exprIfFalse ```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

File: src/components/tabs/CdrTabs.vue [hunk 2]
```diff
@@ -190,7 +192,7 @@ export default {
           }
         }
         if (this.$refs.cdrTabsHeader.children[this.activeTabIndex]
-          && (event.which === 37 || event.which === 39)) {
+          && (event.keyCode === 37 || event.keyCode === 39)) {
```
KrisKnabel: I remember there being a problem with using `event.keyCode` on certain platforms.  I think it was either IE or Safari.

lancetharper: `event.which` is deprecated, and it looks like `event.keyCode` is also not recommended anymore (though browser support looks ok for it https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#Browser_compatibility) 

I'll look into what to do instead 🤔

KrisKnabel:
I pulled the branch and ran it.  The Tab Content Animation(Fly-In/Out) doesn't appear to be working.

src/components/tabs/CdrTabs.vue
cowills: https://medium.com/@uistephen/keyboardevent-key-for-cross-browser-key-press-check-61dbad0a067a this article claims that `var key = event.key || event.keyCode;` should suffice.

cowills:
> So the javascript distributable of the tokens contains `cdrDuration5X: "500ms"`. Being that the delay function desires a unit-less number, should we modify the token, create a new set of unit-less tokens, or cast/transform the value from `"500ms"' to `500' within the function?

i wonder if all the JS tokens should just be unit-less? didn't we do that for breakpoints?

lancetharper:
> i wonder if all the JS tokens should just be unit-less? didn't we do that for breakpoints?

We went from rem to px for breakpoint, but it's something to look at

darintrimillos (dismissed):

darintrimillos (approved):

PR #621, opened 2019-06-03
feat(visibility): display utilities

mhewson:


src/css/main.scss
```diff
@@ -12,7 +12,6 @@
 
 /* Utilities */
 @import './utility/visibility';
-@import './utility/a11y';
 @import './utility/align';
 @import './utility/spacing';
-@import './utility/legacy';
+@import './utility/removeWinter2019';
```
cowills: i think this file should still be named what it is, `deprecatedSpacing` maybe?

cowills: will need to add a file with the deprecated display utilities as well

mhewson: My thought was that we will want a way to time stamp the items to be removed -- this is not just spacing now but display also -- as we deprecate more items it would be nice to know "this batch" is going away in winter and "this other batch" is going away in spring -  I guess that could be a comment in one file but that seems like it could introduce some risk of remove the wrong items?

cowills: we could add the timestamp to the end of the filename. i think ppl migrating these classes will want to be able to lookup these files, so keeping the names meaningful would help that.

cowills:
this is for the fall release right? not part of single package?

src/css/main.scss
mhewson: Ah yea that makes sense

mhewson: something like
deprecated > 
spacing.summer19
display.summer19
ally.summer19

PR #625, opened 2019-06-11
Fix/input v model add watcher

KrisKnabel:
Tried to add unit test for v-model to input value change, but the Vue Test Utils does not support how v-model works at runtime(It does 1 way binding), so there is no way to test this as I know of through Vue Test Utils.  So instead I created an example where we can test it visually in the example.  If you change the master input value, that triggers changes to all the other input's v-models.  This was not working before the watcher was added and is now working.  I feel this is sufficient for now, but we should modify our v-model implementation in the near future to match how Vue recommends.

### DIFF
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -167,6 +167,19 @@ export default {
       };
     },
   },
+  watch: {
+    value(val) {
+      this.newValue = val;
+    },
+    newValue(val) {
+      /**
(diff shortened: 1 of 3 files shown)


cowills (approved):
lgtm. agreed that we can figure out a solid testing strategy later when we refactor stuff to use a common v-model pattern.

mhewson (approved):

PR #629, opened 2019-06-19
Update css output

darintrimillos:
Not very elegant, but I think this does what we want.


lancetharper (commented):
I would vote to not use rollup just for compiling scss. I'd just use the sass/postcss CLI or create a script that uses node-sass/postcss

darintrimillos:
Ok, I'll close this and go that route.

PR #630, opened 2019-06-20
Update css output

darintrimillos:
Using node-sass instead of rollup per @lancetharper suggestion. Much better


cowills:
@darintrimillos  looks like the rollup related commits are still present?

lancetharper (changes requested):
You'll want to run the compiled scss through postcss too to make sure it all gets prefixed, etc. prior to outputting the file

cowills (dismissed):

lancetharper (dismissed):

cowills (approved):

PR #638, opened 2019-06-25
Fix ssr

cowills:
This hacks our package-lock to force rollup-plugin-vue to use vue-runtime-helps@1.0.1 instead of 1.0.0

1.0.1 contains a critical SSR fix: https://github.com/znck/vue-runtime-helpers/commit/02d9dfaf54388128a7eac8013f0af6526f6e8586 

This has been verified on an SSR try server

This hack will be obliterated any time someone runs `npm install` in this repo, so we must address it before the next release. The ideal solution would be to submit a PR/issue to rollup-plugin-vue to update that dependency, then update our rollup-plugin-vue version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rei/cedar",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "REI Cedar Component Library",
   "homepage": "https://rei.github.io/rei-cedar/",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


darintrimillos (approved):

package.json
```diff
@@ -135,6 +135,7 @@
     "tinycolor2": "^1.4.1",
     "vue": "^2.5.22",
     "vue-router": "^3.0.2",
+    "vue-runtime-helpers": "^1.0.1",
```
KrisKnabel, line 138: 👍

KrisKnabel (approved):
Nice work!

PR #639, opened 2019-06-26
fix(ssr): wip styleInjector override

cowills:
we need to pass the browser.js file that is compiled from this branch into styleInjector in rollup-plugin-vue
https://github.com/rei/vue-runtime-helpers/pull/1  (also available in `build/style-injector.js`  in case you don't want to clone+build that repo :p )

I was able to verify this fix by manually editing the contents of `node_modules`, however i can't figure out how to actually use the override API for `styleInjector` that rollup-plugin-vue uses:
https://github.com/vuejs/vue-component-compiler#assemblecompiler-sfccompiler-filename-string-result-descriptorcompileresult-options-assembleoptions-assembleresults 

```
Directly in-lined (default)
Using a global function (styleInjector: 'myComponentNormalizer')
Using an import ({ styleInjector: '~my-component-normalizer' })
```

im guessing "global function" is the solution here


darintrimillos (dismissed):

darintrimillos:
Do not merge until @cowills can review.

darintrimillos (approved):

PR #642, opened 2019-07-08
Breadcrumb update

darintrimillos:
Removed non-ssr-compatible code. Component will truncate by default, can be clicked to reveal full breadcrumb.

### DIFF
--- a/src/components/breadcrumb/CdrBreadcrumb.vue
+++ b/src/components/breadcrumb/CdrBreadcrumb.vue
@@ -73,7 +73,6 @@
 
 import modifier from 'mixinsdir/modifier';
 import breakpoints from 'mixinsdir/breakpoints';
-import debounce from 'lodash/debounce';
 
 export default {
   name: 'CdrBreadcrumb',
--- a/src/components/breadcrumb/CdrBreadcrumb.vue
+++ b/src/components/breadcrumb/CdrBreadcrumb.vue
@@ -42,7 +42,7 @@
       <li
         :class="$style['cdr-breadcrumb__item']"
         v-for="(breadcrumb, index) in items"
-        :key="index"
+        :key="breadcrumb.item.name.replace(/ /g, '-').toLowerCase()"
         v-show="!truncate || (index >= items.length - 2)"
       >
         <a
--- a/src/components/breadcrumb/CdrBreadcrumb.vue
+++ b/src/components/breadcrumb/CdrBreadcrumb.vue
@@ -43,7 +43,7 @@
         :class="$style['cdr-breadcrumb__item']"
         v-for="(breadcrumb, index) in items"
         :key="index"
-        v-show="!truncate || (index >= items.length - 1)"
+        v-show="!truncate || (index >= items.length - 2)"
       >
         <a
           :class="$style['cdr-breadcrumb__link']"
--- a/src/components/breadcrumb/CdrBreadcrumb.vue
+++ b/src/components/breadcrumb/CdrBreadcrumb.vue
@@ -42,7 +42,7 @@
       <li
         :class="$style['cdr-breadcrumb__item']"
         v-for="(breadcrumb, index) in items"
-        :key="breadcrumb.item.name.replace(/ /g, '-').toLowerCase()"
+        :key="breadcrumb.item.id || breadcrumb.item.name.replace(/ /g, '-').toLowerCase()"
         v-show="!truncate || (index >= items.length - 2)"
       >
         <a
--- a/src/components/breadcrumb/CdrBreadcrumb.vue
+++ b/src/components/breadcrumb/CdrBreadcrumb.vue
@@ -1,7 +1,6 @@
 <template>
   <!-- disable lint errors on line length in template -->
   <!-- eslint-disable max-len -->
-  <!-- eslint-disable vue/require-v-for-key -->
 
   <nav
     ref="container"
@@ -43,6 +42,7 @@
       <li
         :class="$style['cdr-breadcrumb__item']"
         v-for="(breadcrumb, index) in items"
+        :key="index"
         v-show="!truncate || (index >= items.length - 1)"
       >
         <a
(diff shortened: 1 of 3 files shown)


src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -43,7 +43,7 @@
       <li
         :class="$style['cdr-breadcrumb__item']"
         v-for="(breadcrumb, index) in items"
-        v-show="!truncate || (index >= items.length - 2)"
```
KrisKnabel: We may need a design decision here, but I don't think we decided to change the minimum number of breadcrumb items.  When I first created the breadcrumbs, I remember design wanting at least 2 breadcrumb items shown even when truncated.  I think we should ask design what they think and even offer an option where we add a prop that controls how many breadcrumb items show when truncated.

darintrimillos: I will follow up on this.

cowills: i agree that this seems like it should be 2, that way the user can move back 1 level without having to expand the breadcrumb

src/components/breadcrumb/CdrBreadcrumb.vue
```diff
@@ -43,6 +42,7 @@
       <li
         :class="$style['cdr-breadcrumb__item']"
         v-for="(breadcrumb, index) in items"
+        :key="index"
```
cowills: using array indices as a key is bad practice, need to use a unique identifier here. (basically, if the array changes then the indices change which makes the keys break.) 
https://rei.slack.com/archives/C61570BL1/p1556748557068600 more details here

KrisKnabel: It looks like Darin was getting a "v-for" warning, so he added the `:key` property.  I had intentionally not defined this and let Vue handle this for performance reasons outlined in the Vue docs.  https://vuejs.org/v2/guide/list.html

`It is recommended to provide a key attribute with v-for whenever possible, unless the iterated DOM content is simple, or you are intentionally relying on the default behavior for performance gains.`

So I feel we should just go back to not defining the `:key` and letting Vue handle it.

cowills: but it's possible to dynamically update the items array, which would break the default index based logic, right? (for example, if someone had a single page app where the breadcrumb updates during navigation) That's why using a unique id as a key would be safer here.

KrisKnabel: Vue under the hood creates a unique key.  We are not modifying the array and thus never using the index or key for modification.  We only use index as essentially a "Read-Only" property for knowing  which two breadcrumb items are at the end of the list.  From my understanding, we would only need to provide unique keys if we did anything where we "Re-order", "Remove", or "Add" to the array.  Especially with SSR this array modification seems like a non-existent use case.

cowills: we aren't modifying the array, but devs easily can. `foo = [1, 2, 3]; <cdr-breadcrumb items="foo">; foo.shift();`

However, if Vue is using a unique key internally then it's fine. I agree that we can just remove the key, ignore the warning, and let vue do it's thing

KrisKnabel: Sorry, I missed the SPA specific part of your concern and assumed REI's use.  The `v-for` is bound to the array as a JS object and not a "Vue" wrapped object.  So users should not get a "bad" reference.  If we had kept `:key="index"` then users could definitely have gotten into trouble by trying to modify the Vue objects directly.

darintrimillos: Converted to use a non-index unique value for :key before seeing this full discussion haha. Thanks for the rundown, y'all.

src/components/breadcrumb/CdrBreadcrumb.vue
darintrimillos: Moved this back to 2 to stay consistent with the previous releases.

src/components/breadcrumb/CdrBreadcrumb.vue
KrisKnabel: I just read: https://forum.vuejs.org/t/simple-clarification-of-when-using-key-with-v-for-is-appropriate-and-why/28966/8  It may be best to define the `:key` just as an extra precaution.  I think we could add the key to look for a provided `id` before just using the `name`

KrisKnabel: I propose we use lodash.uniqueId.  So `:key="item.id || _.uniqueId(item.name)"`

KrisKnabel: Sounds like Vue already has a duplicate key warning, so we don't need to use `uniqueId`, but I now have a concern with the items data structure.  Seems like we have two different formats on the doc site.  
image

cowills: looks like the first example is correct. However the second example is clearly better. But fixing that would be a breaking change so it would have to wait for the fall release

cowills: i think `:key="breadcrumb.item.name.replace(/ /g, '-').toLowerCase()"` makes sense as a key, as it will always be unique . looks like lodash uniqueid would basically do the same thing, except it also adds a random number to the end

KrisKnabel: I still think we should bind to a provided `id` if its there.  `:key="breadcrumb.item.id || breadcrumb.item.name.replace(/ /g, '-').toLowerCase()"`

KrisKnabel (approved):

src/components/breadcrumb/CdrBreadcrumb.vue
darintrimillos: Fixed.

PR #649, opened 2019-07-18
Update/accordion

darintrimillos:
This isn't done yet, but I wanted to get some dev feedback on how this is working. 

I removed the internal state tracking; it now updates when the show prop is updated from the parent. There's a watch set on the show prop, which triggers the animation-related stuff on update. I will add another example in here that has the open one/close others logic.


cowills:
outright deleting accordion would be a big breaking change, is that necessary?

src/components/accordion/CdrAccordionItem.vue
```diff
@@ -2,22 +2,24 @@
...
     :class="[modifierClass, styleClass, focusedClass]"
     :id="`${id}-accordion-item`"
-    :ref="`accordion-${id}`"
+    :ref="`accordion-container`"
   >
     <button
       :class="$style['cdr-accordion-item__button']"
       :id="id"
-      @click="toggle($event)"
+      @click="$emit('accordion-item-toggle', $event);"
```
cowills: i think this naming convention for component events makes sense, as long as we're consistent about the use of the word "toggle" and so on

lancetharper:
> outright deleting accordion would be a big breaking change, is that necessary?

Yes, and if we do that I'd vote we remove `accordion-item` and keep `accordion`

darintrimillos:
> outright deleting accordion would be a big breaking change, is that necessary?

I do think we want to delete it. Right now it's just wrapper that passes a few values into `cdr-accordion-item`. I have no problem with renaming `accordion-item` to simply `accordion`.

Is it not a big breaking change even if we keep it? Since it's now relying on a prop update to toggle, consumers would need to add code to update that prop when the event fires anyway. Another option is we keep `accordion` and build that open-one-close-the-others functionality into it so it works outta the box. I'd rather just have a simpler single component.

cowills:
I think turning it into just accordion makes sense, we just need to be aware of how many breaking changes we are making, ensure they are documented, etc.

PR #650, opened 2019-07-18
map variables for dark/light CTA

cowills:
did you know that the CTA component has "light" and "dark" styles which have nothing to do with what we call "lightmode" and "darkmode"? 😆 🤣 😢 😭 😪 

the CSS for CTA was written to make "brand" the default mode, but the vue component sets the default mode to "dark" so that brand default CSS never executes ever. Fun stuff. I just deleted that set of base color/background/box-shadow/fill variables and will update the comp-vars docs to note that you need `base + exactly one mode` for it to work.


lancetharper (approved):
There's been a lot of shifting sands in the life of these early components 😅

KrisKnabel (approved):

fesnaqvi (approved):

PR #652, opened 2019-07-19
tree shaking

cowills:
WIP

The per-icon components are a huge pain here :~( 

Row and Col being in a folder called "grid" is also a pain :~(

In general, there's a lot of internal re-org we could do now that we're off lerna to make the build simpler. (could definitely be done as part of the test/example re-org)

I disabled the SSR builds because FEBs doesn't even support that. ALso disabled the big ESM build and the single file CJS builds because they make no sense.

Consumers will need to set this in their webpack.overrides.conf.js so FEBS can build the modules properly, it looks like this:

```
const dir = __dirname
module.exports = {
  module: {
    rules: [{
      test: /\.js$/, // TODO: target cedar only? hrmmm.... export this as a loader somehow?
      use: {
        loader: 'babel-loader',
        query: {
          plugins: [
            [require('babel-plugin-transform-imports'),  {
              '@rei/cedar': {
                transform: (importName, matches) => `${dir}/node_modules/@rei/cedar/dist/${importName.toLowerCase()}/${importName}.esm.js`,
                preventFullImport: true,
              }
            }]
          ]
        }
      }
    }]
  }
}

```

That transform just turns this:

`import {CdrText, CdrButton} from '@rei/cedar'`

into this:

```
import CdrText from 'BUILD_DIR/node_modules/@rei/cedar/cdrtext/CdrText.esm.js'
import CdrButton from 'BUILD_DIR/node_modules/@rei/cedar/cdrbutton/CdrButton.esm.js'
```

This would allow fedpack components to be agnostic about CJS vs. ESM and let the consuming micro-site decide how to build it.


cowills:
found a waaaaaaaay simpler solution 🤦‍♀ 😆

PR #653, opened 2019-07-29
fix utilities.scss build

cowills:
fixes the utilities.scss imports to load the new utility files. (also updates two instances of `cdr-sr-only` to match the new name for that class.

side note: we should probably not use utility classes in the components, since it means cedar consumers then have to load the utilities even if they never use them

### DIFF
--- a/src/components/dataTable/CdrDataTable.vue
+++ b/src/components/dataTable/CdrDataTable.vue
@@ -28,7 +28,7 @@
         :id="id ? id : null"
       >
         <caption
-          class="cdr-sr-only"
+          class="cdr-display-sr-only"
           v-if="caption"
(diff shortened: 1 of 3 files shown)


lancetharper (dismissed):
Should we update those to also be a mixin so we can use them in components as well as in the utility class?

cowills:
> Should we update those to also be a mixin so we can use them in components as well as in the utility class?

That's one possible solution. I don't think we really need to solve this problem until/unless we move utilities to it's own thing. For now just avoiding use of utilities in components seems fine

KrisKnabel (dismissed):

darintrimillos (approved):

PR #655, opened 2019-07-31
Accordion updates

darintrimillos:
Got design approval from Fes, so we should be ready to go here.


src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,9 +1,109 @@
...
+    //     },
+    //     slots: {
+    //       default: 'this is some default slot text. it should take up some space'
+    //     }
+    //   });
+
+    //   expect(wrapper.vm.maxHeight).toBe(0);
+
+    //   // this is throwing a console error, but shouldn't be. Covered in this thread
+    //   // https://github.com/vuejs/vue-test-utils/issues/631
```
cowills: this issue looks like it was resolved already? the PR they reference as the solution was merged

cowills: also shouldn't this be `setProps` since opened isn't in `data()`?

lancetharper:
Looks like `chromedriver` is outdated with the recent Chrome update

src/components/accordion/CdrAccordion.vue
```diff
@@ -1,56 +1,139 @@
...
+
+      if (this.borderAligned) {
+        styles.push(this.modifyClassName(this.baseClass, 'border-aligned'));
+      }
+
+      return styles.join(' ');
+    },
+    focusedClass() {
+      return this.focused ? this.modifyClassName(this.baseClass, 'focused') : null;
+    },
```
KrisKnabel, line 126: This `focused` classname logic at first glance seems unnecessary as you could use the `:focus` pseudo-class in the stylesheet, but I looked in the .scss file and it looks as if you have something going on with changing the z-index.  Is this old logic that could possibly be removed?  If not, maybe add a comment as to why the z-index change.

darintrimillos: This is to fix a Chrome-only visual issue. I added a comment into the stylesheet to explain why it's there.

src/components/accordion/__tests__/CdrAccordion.spec.js
darintrimillos: Yes, it should have been `setProps`. We actually don't need this anymore as it was testing some stuff that was refactored out.

darintrimillos: I added an additional assertion and a comment.

cowills:
@darintrimillos before moving onto other work, can you also update the fall release notes in cedar-docs to document the breaking changes here?

darintrimillos:
> @darintrimillos before moving onto other work, can you also update the fall release notes in cedar-docs to document the breaking changes here?

Yeah, I'll do that.

cowills (approved):
lgtm, animations are much snappier now 👍

PR #656, opened 2019-07-31
fix(cta, vars): create light-elevated mixin

cowills:
cta-style="light" and modifier="elevated" results in this special case :<


lancetharper (dismissed):

lancetharper:
Looks like the recent Chrome update is breaking a test

KrisKnabel (approved):

PR #657, opened 2019-08-01
feat(link): `CdrLink` emits a click event

lancetharper:


cowills (dismissed):
:+1: lgtm

src/components/link/CdrLink.vue
```diff
@@ -5,6 +5,7 @@
     :target="target"
     :rel="computedRel"
     :href="tag === 'a' ? href : null /* don't include the href attribute if not an <a> */"
+    @click="$emit('click', $event)"
   >
```
KrisKnabel, line 9: I have a feeling it might be ideal to name our click event something different than the default `click` in order to be more deliberate about what actual event the consumer is listening to.  I remember having strange behavior depending on where I was listening for the event.  Should we maybe name it `linkClick`?

cowills: i think click is fine since we are passing along the event? what do other vue libraries do? I don't think having a unique "click" event name for every component is a good idea

lancetharper: I actually prefer just using `click` since it's how Vue works. We aren't doing anything different with it and it should work identically anyhow.

I'm wondering if we should just just pass all listeners through...then we don't have to handle any and all cases since we don't really care about the click/other events anyway.

https://vuejs.org/v2/api/#vm-listeners

lancetharper: Yeah, I'm going to change to that since we're basically making a "transparent wrapper component" anyway. It's probably a pattern we should follow elsewhere too unless our component specifically needs to know about an event and do something.

cowills: that seems like it might also solve the pagination vue router click event issue Joe reported as well? smth like `<CdrPagination @click.prevent="handlePaginationClick" />`. Would just be on consumers to check the `event.target.href` and handle it accordingly

cowills: > unless our component specifically needs to know about an event and do something.

even then, presumably the component could internally catch the event and then re-emit the event after transforming it? or just let it bubble up i guess? will have to play around more with this.

lancetharper: Yes, my thought exactly. We can listen to an event and emit the same event...then things would work more how you'd expect Vue things to work without having to know an explicit "event API" of what is/isn't supported for every component.

We should still call out that events are passed through in the component docs though so people at least know they are.

KrisKnabel (approved):

PR #659, opened 2019-08-08
v-model/listeners consistency

cowills:
- updating anything that has a v-model to make it:
--  use `value` as the v-model attr
--  emit an `input` event when that value changes
- using the $listeners pattern wherever possible to replace previously hardcoded bindings

radio, pagination, and select were the only components that deviated from the standard v-model pattern. all three will have simple breaking changes for JSX users (they'll just need to rename the prop they bound to.) Aliasing props does not seem to be a thing in Vue, so not sure there is a good way to gracefully deprecate those without possibly causing a bigger mess.

docs updates: https://github.com/rei/rei-cedar-docs/pull/390

### DIFF
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -153,20 +153,20 @@ export default {
         [this.$style['cdr-input-wrap']]: true,
       };
     },
-    inputListeners: function () {//
+    inputListeners() {
       // https://vuejs.org/v2/guide/components-custom-events.html#Binding-Native-Events-to-Components
       // handles conflict between v-model and v-on="$listeners"
-      var vm = this;
+      const vm = this;
       return Object.assign(
         {},
         this.$listeners,
         {
-          input: function (event) {
-            vm.$emit('input', event.target.value)
-          }
-        }
+          input(event) {
+            vm.$emit('input', event.target.value);
+          },
+        },
       );
-    }
+    },
   },
 };
 </script>
--- a/src/components/input/CdrInput.vue
+++ b/src/components/input/CdrInput.vue
@@ -26,13 +26,9 @@
         :rows="[rows]"
         :class="[inputClass, sizeClass, space]"
         v-bind="$attrs"
+        v-on="$listeners"
         v-model="newValue"
         :id="inputId"
-        @blur="onBlur"
-        @input="onInput"
-        @focus="onFocus"
-        @paste="onPaste"
-        @keydown="onKeydown"
         :disabled="disabled"
         :required="required"
         :aria-label="hideLabel ? label : null"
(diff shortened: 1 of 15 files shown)


src/components/input/CdrInput.vue
```diff
@@ -166,61 +153,19 @@ export default {
...
+      const vm = this;
+      return Object.assign(
+        {},
+        this.$listeners,
+        {
+          input(event) {
+            vm.$emit('input', event.target.value);
+          },
+        },
+      );
```
KrisKnabel, line 168: This is so much better!  💯

src/components/input/__tests__/CdrInput.spec.js
```diff
@@ -201,58 +202,86 @@ describe('CdrInput.vue', () => {
...
+    const wrapper = shallowMount(CdrInput, {
+      propsData: {
+        label: 'test',
+        value: 'bar'
+      },
+    });
+    const input = wrapper.find({ ref: 'input' });
+    wrapper.setProps({value: ''});
+    expect(input.element.value).toBe('');
+  });
```
KrisKnabel, line 286: Clever  👍

src/components/pagination/CdrPagination.vue
```diff
@@ -131,14 +127,13 @@ export default {
         return result;
       },
     },
-    /** @ignore used for binding v-model */
-    currentPage: {
+    /** @ignore used for binding v-model, represents the current page */
```
KrisKnabel, line 130: Do we still need the @ignore?  I removed it locally and I don't see any errors.

cowills: since we aren't auto-generating API docs anymore it's not necessary. We could probably make a separate chore task to remove all of those

src/components/input/CdrInput.vue
cowills: yeah, i think we can do something similar to checkbox/radio/select as well. gonna save that for a future PR though since this is already pretty big.

src/components/pagination/CdrPagination.vue
```diff
@@ -221,8 +216,7 @@ export default {
     },
   },
   watch: {
-    currentPage(v) {
-      this.localCurrent = v;
+    value() {
       this.currentUrl = this.pages[this.currentIdx].url;
     },
   },
```
KrisKnabel, line 222: I think we can get rid of the watcher on `value` and update `currentUrl` in the `select` method that is bound to the `change` event.  As this is a large PR, we could worry about the change in a later chore/refactor PR.

KrisKnabel (approved):

lancetharper (approved):
🎉

src/components/pagination/CdrPagination.vue
cowills: this watcher has to run for any value update, not just the select. and it can't be a computed prop since it's passed as the v-model to the select. 
agreed that there's definitely a better pattern to use here though :+1:

PR #662, opened 2019-08-14
Accordion bug fix

darintrimillos:
I found a bug in accordion when updating the docs site with the alpha @cowills made. I've fixed that bug in this PR, and tested that code in the docs site nav.

I'd like to get this merged and create a new alpha for docs site, which will then unblock my PR there.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rei/cedar",
-  "version": "2.1.0-alpha.0",
+  "version": "3.0.0-accordion.alpha.1",
```
cowills: can you just name this alpha.3? i don't think it makes sense to issue named alphas off of next

darintrimillos: Whoops, yeah just did that to make an npm package yesterday so it wouldn't get the alpha tag. I'll fix that.

cowills (approved):

PR #665, opened 2019-08-27
Feature/type

mhewson:


static/cdr-fonts.css
```diff
@@ -1,6 +1,21 @@
 /*
 * This is the unminified version of cedar-fonts.min.css
 */
+@font-face {
```
cowills, line 1: can we drop the legacy fonts here? or possibly just delete the inline base64 fonts from the old ones? they add a huge amount to consumer bundles

static/cdr-fonts.min.css
```diff
@@ -1 +1,17 @@
+@font-face {
```
cowills: should actually minify this if it's `min.css`

src/css/deprecated/fall2019/typeography.scss
```diff
@@ -0,0 +1,266 @@
+/* subheading */
```
cowills, line 1: we can just delete these, these mixins don't get exported

cowills: related to this change: we should probably remove typography from the component variables entirely and just have people use utility classes for text.

static/cdr-fonts.css
mhewson: I was going to ask about that at dev review! I think it makes sense as long as that still allows us to support the deprecated type settings . -- seems like removing the inline should be don't either way though

static/cdr-fonts.min.css
mhewson: yep getting there

src/css/deprecated/fall2019/typeography.scss
mhewson: They are used in the deprecated classes though - I guess they could be deconstructed

mhewson: as in 
```
  &--subheading {
    @include cdr-text-subheading-mixin;
  }
```

src/components/accordion/examples/Accordion.vue
```diff
@@ -2,15 +2,15 @@
   <div class="accordion-container">
     <cdr-text
       tag="h2"
-      modifier="heading-small"
+      modifier="heading-600 heading-700@md heading-700@lg"
```
KrisKnabel, line 5: Just checking, but this looks like it got changed to the `heading-medium` new values instead of the `heading-small` values.  Was that the intent?

mhewson: Yes I took that to be a mistake to have the H2 at a smaller size than the h3's but this can be changed if needed

KrisKnabel (approved):

PR #672, opened 2019-09-03
fix(fullwidth): the fullwidth prop was used as a string rather than a…

mhewson:
… bool on some examples

also took the oportunity to do an audid update and apply the space prop to the buttons

### DIFF
--- a/src/css/generic/_color.scss
+++ b/src/css/generic/_color.scss
@@ -28,10 +28,3 @@
   background-color: $cdr-color-background-darker;
 }
 
-//--------------------------------------------------------------------------------------------------
-// Foreground color
-//--------------------------------------------------------------------------------------------------
-.cdr-color-foreground-secondary{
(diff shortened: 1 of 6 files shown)


KrisKnabel (dismissed):

src/css/generic/_color.scss
```diff
@@ -26,3 +27,11 @@
 .cdr-bg--darker {
   background-color: $cdr-color-background-darker;
 }
+
+//--------------------------------------------------------------------------------------------------
+// Foreground color
+//--------------------------------------------------------------------------------------------------
+.cdr-color-foreground-secondary{
```
cowills: ??? is this supposed to be in this PR?

mhewson: thats for type - I'll move it to it's own pr for clarity

KrisKnabel (approved):

PR #679, opened 2019-09-11
fix(tests): temporary fix for chromedriver in CI environment

cowills:
this will allow us to unblock the CI check. if 77 isn't out before we want to release then we can rollback to 76, doesn't matter much since this is just a devDep.

https://github.com/giggio/node-chromedriver/pull/227 the branch/PR is just 2 version bumps so LGTM

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ addons:
   chrome: stable
 dist: xenial
 before_install:
-- if [ "$TRAVIS_OS_NAME" == "linux" ]; then export CHROME_BIN=/usr/bin/google-chrome;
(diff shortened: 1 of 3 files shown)


lancetharper (dismissed):

lancetharper:
I'm not sure if it's a chrome/chromedriver mismatch or what, but it still doesn't want to run correctly.

The chromedriver patch says `77.0.3865.40` but the installed chrome version is `77.0.3865.75-1` which I assume should match

Edit: Nvm. Got it working 🎉

lancetharper (approved):

PR #681, opened 2019-09-17
fix(dev): make dev env work in ie11

cowills:
this is ugly but it fixes it :+1:

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -7,14 +7,29 @@ module.exports = function (api) {
 
   api.cache(true);
 
+  // TODO: refactor to use babel `env` API here instead
+
+  const targets = env === 'test' ?
+  {
+    node: 'current',
+  } :
+  {
+    browsers: [
+      'Chrome >= 70',
+      'Firefox > 64',
+      'iOS > 11',
+      'Safari >= 9',
+      'Explorer >= 11',
+      'Edge >= 15',
+    ],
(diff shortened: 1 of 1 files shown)


babel.config.js
```diff
@@ -7,14 +7,29 @@ module.exports = function (api) {
...
 
+  // TODO: refactor to use babel `env` API here instead
+
+  const targets = env === 'test' ?
+  {
+    node: 'current',
+  } :
+  {
+    browsers: [
+      'Chrome >= 70',
```
cowills, line 18: this is what febs currently compiles to: https://github.com/rei/front-end-build-configs/blob/master/application/webpack.base.conf.js#L114-L121

there is also a package named smth like "@rei/browserslist-config" on the private registry which has a similar browser matrix. (though that one is based on some dubious metrics and needs to be updated)

long term we should make febs and cedar pull from the same thing

babel.config.js
```diff
@@ -7,14 +7,29 @@ module.exports = function (api) {
 
   api.cache(true);
 
+  // TODO: refactor to use babel `env` API here instead
```
cowills, line 10: babel lets you set multiple configs that get switched based on the value of `BABEL_ENV`, which would clean up this mess and make it more understandable. we would need to re-factor other parts of the build to make use of it though.

KrisKnabel (approved):

PR #683, opened 2019-09-24
Update icon

lancetharper:
- CdrIcon now accepts full svg markup in slot
  - Attributes/classes/event listeners are all preserved and our defaults are preserved (viewBox, etc.)

This feature allows us to accept raw svg markup effectively eliminating the need for individual icon components and making it easier to use with cedar compliant but "non-cedar" (not in our icon repo) svg icons


src/components/icon/CdrIcon.jsx
```diff
@@ -30,11 +31,28 @@ export default {
...
+        slotDataObj = vNode.data;
+        // remove wrapping svg from slot but keep its contents
+        this.$slots.default = vNode.children;
+      }
+    }
+    
     return (<svg
-      xmlns="http://www.w3.org/2000/svg"
-      viewBox="0 0 24 24"
+      {...merge({}, slotDataObj, defaultDataObj)}
```
cowills: i think you can do this instead of using merge here:
``` 
{...slotDataObj}
{...defaultDataObj}
```

JSX just builds an object structure, so if both Obj's have a viewBox the one in defaultData would override slotData because its added last.

lancetharper: Will that recursively merge? Let me try it out

cowills: oh right, it would need to actually merge the `attrs` object and it won't do that in this case

cowills (dismissed):

src/components/icon/CdrIcon.jsx
lancetharper: It looks like that does work! Good to know. Let's just roll with that and keep the code smaller

cowills (approved):

PR #686, opened 2019-09-27
Feat/select input consistency

KrisKnabel:
Select Updates to conform to Input

To view this build go to: http://rei-cedar-components-demo.s3-website-us-west-2.amazonaws.com/#/selects

Known Issues:

Firefox: 
  - Dotted outline does not match up to border
  - Prompt styling affects options before selection is made, then is normal

Internet Explorer/Edge
  - Same prompt styling issue as Firefox
  - FIXED ~~V-Model not updating selection(Fixing this now)~~


KrisKnabel:
Fixed IE v-model issue.

KrisKnabel:
Fixed Linter space complaint.  This should be ready for review.

src/components/select/styles/vars/CdrSelect.vars.scss
```diff
@@ -1,69 +1,207 @@
...
+
+@mixin cdr-select-helper-text-mixin() {
+  @include cdr-select-helper-text-text-mixin;
+
+  color: $cdr-select-helper-text-color;
+  padding-left: $cdr-select-helper-text-pl;
+  padding-top: $cdr-select-helper-text-pt;
+}
+
+// $cdr-select-base-font-size: 16px;
```
lancetharper: Is there any reason to keep all of this commented out stuff?

src/components/select/CdrSelect.jsx
```diff
@@ -65,98 +75,73 @@ export default {
...
+      const vm = this;
+      return Object.assign(
+        {},
+        this.$listeners,
+        {
+          input(event) {
+            if (vm.multiple) {
+              const optArr = toArray(event.target.options);
+              const selected = optArr.filter(o => o.selected === true).map(o => o.value);
+              vm.$emit('input', selected);
```
lancetharper: missing emitting the event in a couple places (and unit tests for these new events)

lancetharper (changes requested):

src/components/select/styles/vars/CdrSelect.vars.scss
KrisKnabel: Not anymore, I will remove these comments.

src/components/select/CdrSelect.jsx
KrisKnabel: Good callout on the missing unit tests.  I will add some.

KrisKnabel: What places/scenarios do you think the event emitting is missing?

src/components/select/CdrSelect.jsx
```diff
@@ -65,98 +75,73 @@ export default {
...
+
+              // Deprecated Event
+              vm.$emit('change', selected, event);
```
lancetharper, line 98: @KrisKnabel Like here. Emit the `value` (selected), and the `event`

src/components/select/CdrSelect.jsx
KrisKnabel: I think what you are asking is to pass through the original event object.  I looked at how Cole refactored the Input component, and she did not attach the event object.  I then looked at the Vue docs: https://vuejs.org/v2/guide/components-custom-events.html  they also don't attach the event object.  I guess adding the original event object won't hurt and could be useful.

src/components/select/CdrSelect.jsx
KrisKnabel: Attached original event object

src/components/select/CdrSelect.jsx
lancetharper: Yeah it's probably not a big deal, I just think we're just better off having it in there now rather than someone discovering they need/want it later and us having to make an update (or push it to the next release)

src/components/select/CdrSelect.jsx
lancetharper: Sweet, don't forget to update the emitted events docs to include it! 👍

lancetharper (approved):

PR #689, opened 2019-10-08
Restore unit test coverage

lancetharper:
- No longer need to use built cedar.js (test against source files)
- Code coverage per file ~(and requires 100% lines covered)~
- some cleanup of old files and now unused packages
- ~update unit tests (or ignore  if unimportant) to restore 100% line coverage~
- update unit tests with jest snapshots
- add codecov support (using rei-cedar github account)

There is probably additional cleanup that can be done in babel and rollup since we just use `jest` now and don't use the `BABEL_ENV` stuff anymore but the e2e tests still set `NODE_ENV` to 'test' so I didn't want to touch those right now.


src/components/breadcrumb/CdrBreadcrumb.jsx
```diff
@@ -16,13 +16,16 @@ export default {
     items: {
       type: Array,
       default: () => [],
+      /* istanbul ignore next */
```
cowills: what are these ignores about?

lancetharper: Ignoring the validator function. We can keep them if wanted but since they don't actually do anything for the functionality of a component I didn't think it mattered to test them.

cowills: i think it would be better to lower the coverage reqs than ignore lines of code. (also...shouldn't this validator function run at least once in one of the tests? like if you instantiate a breadcrumb with an items array in one of the tests that should make this line "covered")

mhewson (approved):

src/components/accordion/__tests__/CdrAccordion.spec.js
```diff
@@ -1,6 +1,5 @@
-import { shallowMount } from '@vue/test-utils';
-import { CdrAccordion } from 'distdir/cedar.js';
-import sinon from 'sinon';
+import { shallowMount, mount } from '@vue/test-utils';
```
KrisKnabel, line 1: I'm assuming the new snapshot tests need { mount ) imported?  Seems like all the tests are using shallowMount instead.

lancetharper: I don't think _all_ of them do, but `mount` will render children and stuff so it's safer to just use it everywhere to make sure that the snapshot is complete.

cowills (approved):

PR #693, opened 2019-10-11
fix(select): fix caret icon to show disabled styling

KrisKnabel:
Caret icon is not showing disabled, this change fixes the caret icon to display properly when
disabled.  Pointed out by the Cedar Designers and they requested the fix.


src/components/select/styles/vars/CdrSelect.vars.scss
```diff
@@ -167,3 +167,19 @@ $cdr-select-helper-text-pt: $cdr-space-quarter-x;
   padding-left: $cdr-select-helper-text-pl;
   padding-top: $cdr-select-helper-text-pt;
 }
+
+/* caret */
+@mixin cdr-select-base-caret-mixin() {
+  fill: $cdr-color-icon-primary-lightmode;
```
cowills: you need to create variables for all these properties if you're going to map these like this. It's fine to just leave this CSS un-mapped instead.

lancetharper (dismissed):

cowills (changes requested):

src/components/select/styles/vars/CdrSelect.vars.scss
KrisKnabel: When you say variables, are you referring to Cedar Component Variables?

cowills: the `vars.scss` files are the component variables. if you aren't going to map each of these values to a variable then you should just keep these changes in the main scss file for select

KrisKnabel: I see what you mean now.  Sorry for the confusion.  I was looking at the code above the comment, not below, and wondering if you were referring to running `npm run build:variables`.

KrisKnabel: Updated the select caret style properties to map to SCSS variables.

KrisKnabel:
Added Unit Test to cover Cedar Select Caret Disabled state.

KrisKnabel:
Closing as PR 696 resolves the need for this.

PR #695, opened 2019-10-15
Re fix rating 2

cowills:
3.0.0 introduced a bug where passing the number 0 as count caused an unstyled "0" to appear in the UI. (0 is also the default value for count)
3.0.1 fixed this but introduced a new bug if user did not pass a count at all (default was still 0, so it would show "0 reviews" even if you passed nothing)
this change restores the behavior from 2.0.2 (i.e, if you don't pass anything in as count, then that text does not show)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rei/cedar",
-  "version": "3.0.1",
+  "version": "3.0.2",
   "description": "REI Cedar Component Library",
   "homepage": "https://rei.github.io/rei-cedar/",
   "license": "MIT",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@rei/cedar",
-  "version": "3.0.1",
+  "version": "3.0.2",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rei/cedar",
-  "version": "3.0.1",
+  "version": "3.0.2",
```
KrisKnabel, line 3: Any Chance we can also get the Select Caret Fix out with this release/version as well?

KrisKnabel (approved):

lancetharper (approved):

package.json
cowills: not unless the select caret fix is a new bug in 3.0.0

KrisKnabel: It was a bug discovered by design when they saw the cedar select component on the doc site that we decided to go ahead and release with as Hewson did not want to merge in the fix the day of the release and said it would be a fast follow.  I know design wants the fix out there as the Cedar Select Component currently displays incorrectly in its disabled state.

PR #697, opened 2019-10-15
fix(icon): update icon build to properly bind props

cowills:
this fixes the accordion compact icon issue, which was definitely introduced in 3.0.0

cdr-icon was working correctly with size, but the single icon components were not due to a JSX syntax error in that code. `{...{props: this.$props} }` is the correct way to pass props through to a child component in vue JSX.


KrisKnabel (approved):

lancetharper (approved):
Ah, of course! It needs that full object

PR #699, opened 2019-10-16
Backstop update

cowills:
backstop process notes: https://gist.github.com/cowills/e22d20888eaeebb6c2eef30e44dae024

- update to latest backstop
- switch driver from `chromy to `puppeteer`
- fix tests that were throwing errors

tokens: https://github.com/rei/rei-cedar-tokens/pull/67
variables: https://github.com/rei/rei-cedar-component-variables/pull/12


backstop.js
```diff
@@ -1,17 +1,39 @@
...
 const scenarioDefaults = {
   url: 'http://localhost:3000',
   misMatchThreshold: 0.1,
-  readyEvent: null,
+  readyEvent: 'BACKSTOP_READY',
+  delay: 500,
   onReadyScript: 'onReady.js',
 };
 
+// TODO: use breakpoint tokens for viewport widths
```
cowills, line 13: makes it so you can pass `responsive: true` in a backstop scenario to make it test all 4 breakpoints.
otherwise, it just tests at the largest one.
this makes the tests run more consistently, reduces the number of screenshots we need to deal with, etc.

src/router.js
```diff
@@ -49,8 +58,15 @@ const routes = [
   { path: '/selects', name: 'Selects', component: Selects },
   { path: '/tabs', name: 'Tabs', component: Tabs },
   { path: '/texts', name: 'Texts', component: Texts },
-  { path: '/space', name: 'Space', component: Space },
-  { path: '/utilities', name: 'Utilities', component: Utilities },
+  { path: '/default-buttons', name: 'Default Buttons', component: DefaultButtons },
```
cowills: this is maybe a lil funky... 
some of our example pages are too big or have too much going on which causes backstop to mess up at certain viewports. To get around this, i added routes to individual component examples and made additional scenarios targeting them specifically (see buttons and utilities backstops).

i did make it so these routes do not get rendered at the top of the page, they're just available as paths for backstop to look at

cowills:
A few things that will become follow up tasks:

- audit existing backstop targets to verify hover/focus and responsiveness are adequately covered, and split up existing scenarios/examples as needed 
- generate initial set of reference snapshots, verify they are good
- set up a mechanism for sharing snapshots among dev team
- put kitchen sink on gh-pages, ensure it renders in major browsers

src/components/Utilities/utilities.backstop.js
```diff
@@ -1,15 +1,44 @@
...
+    responsive: true,
+    selectors: [
       '[data-backstop="utility-container"]',
     ],
-    focusSelectors: [
-      '[data-backstop="a11y-skip-link"] > a',
+  },
+  {
+    url: 'http://localhost:3000/#/margin-spacing-utilities',
+    label: 'Margin Spacing Utilities',
```
cowills, line 28: for whatever reason backstop or puppeteer or smth just freaks out about margin-top and padding-top it seems? moving these 2 examples to their own pages and capturing the entire document rather than a specific selector seems to make it work consistently

backstop.js
```diff
@@ -41,13 +63,18 @@ defs.forEach((def) => {
   if (Object.prototype.hasOwnProperty.call(currDef, 'hoverSelectors')) {
     const { hoverSelectors } = currDef;
     delete currDef.hoverSelectors;
-    splitScenario(currDef, 'hover', hoverSelectors, { hoverWait: 1000 });
```
lancetharper, line 44: Is this preserved somewhere? I didn't see it. The reason to have a configurable delay after hover/focus was so things with transitions/etc. could finish before we screenshot (like button background color)

cowills: puppeteer handles hover/focus better than chromy, so it's no longer necessary (at least for our current styles/tests.)

I added back an option for the hover and focus tests so you can add `wait: 1000` in a scenario if it needs to wait after the interaction

lancetharper (approved):

PR #704, opened 2019-10-24
Backstop round two

cowills:
- adding responsive scenarios for every component that has any of the `-mq` helpers in it's scss
- adding hover/focus scenarios for every component that has :hover or :focus in it's scss
- updating docs


src/components/checkbox/CdrCheckbox.backstop.js
```diff
@@ -1,12 +1,18 @@
...
-    selectors: [
-      '[data-backstop="checkboxes"]',
-    ],
-    hoverSelectors: [
-      '[data-backstop="checkbox-checked"]',
-    ],
+    label: 'Checkbox responsive',
+    responsive: true,
   },
+  // TODO: need method for focus/hovering one element but snapshotting it's parent
```
cowills, line 7: since we use the "input hidden behind wrapper" pattern for checkbox and radio, if we take a snapshot of the element that receives focus we just get a tiny slice of the label.

We need some way to say "focus/hover/click this selector but snapshot this other selector"

This may be useful in other scenarios, for example if you could render 2 buttons side by side, focus one of them, but snapshot both of them it would make it much easier to see if the focused button actually updated.

lancetharper (approved):

PR #705, opened 2019-10-25
backstop: add ability to target one element but capture another

cowills:
previously, backstop was set up so hoverSelectors and focusSelectors both interacted with and captured the same element. This causes a bunch of issues:
- hard to tell if an element is focused/hovered or not without seeing it next to an non-focused/hovered version
- lots of cedar components pass through their `$attrs` to an internal element, which includes `data-backstop`...so if you attach that to a cedar component instead of a wrapper element you end up missing most of the component (for example, the label elements around input and select)
- some components like checkbox and radio use a hidden native element behind a UI wrapper, making it difficult to craft a selector that works for both interaction and capturing

This change lets you pass objects to either of the `___Selectors` arrays in a scenario, specifying a separate `target` and `capture` selector. passing strings will continue to behave as before.

lets us do neat things like this:
Screen Shot 2019-10-25 at 10 39 33 AM

### DIFF
--- a/backstop.js
+++ b/backstop.js
@@ -44,8 +44,8 @@ function splitScenario(def, type = '', selectorsArr, extraOptions = {}) {
   const scenario = Object.assign({}, extraOptions, def);
   selectorsArr.forEach((s, idx) => {
     scenario.label = `${def.label} ${type}${idx}`;
-    scenario[`${type}Selector`] = true;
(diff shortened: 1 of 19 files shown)


lancetharper (approved):
Rad, I like that backstop is extensible like this.

KrisKnabel (approved):

PR #707, opened 2019-10-29
export per-component css

cowills:
## Description

See https://git.rei.com/projects/CLIMB/repos/climbers-site/pull-requests/103/overview on bitbucket for febs/climbers related updates and an example of how this is used.

### Changes

- Creates an additional CSS output in `/dist/style` containing a "decomposed" version of `cedar.css`. 
- Moves reset and utilities into that folder. 
- Creates a `CdrText.scss` file that just imports the utilities. 
- Also creates a `/dist/cdr-full.css` file which simply imports everything inside `/dist/style`. This will allow consumers to still just "import everything" if they want, while still getting the benefits of de-duping any dependencies. 

### Testing

- We will be able to "test" that postcss imports are working for a version of cedar by loading it into climber-details-page and climbers-site and verifying that the page renders correctly. 
- FEBS also has unit tests added to it which verifies that it is handling postcss imports correctly. 
- The docs site can be used to test the old "cedar.css" output. 

tested the following use cases manually against climbers-site and search:

- 2 dependencies each import CdrButton@3.0.0 => one copy of CdrButton is bundled
- 1 dependency imports CdrButton@3.0.0 and another imports CdrButton@4.0.0 => 1 copy of each version is bundled
- 1 dependency imports CdrButton@3.0.0 and another imports cdr-full@3.0.0 => one copy of CdrButton is bundled

TODO: should we also decompose the utilities? utils are 100kb i.e, 1/3 of cedar.
TODO: additional testing needed?

### DIFF
--- a/build/extract-css.js
+++ b/build/extract-css.js
@@ -1,15 +1,41 @@
 const sass = require('node-sass');
 const tildeImporter = require('node-sass-package-importer');
 const fs = require('fs-extra');
+const glob = require("glob");
 const chalk = require('chalk');
 const postcss = require('postcss');
 const postcssrc = require('postcss-load-config');
+const postcssModules = require('postcss-modules');
+const packageJson = require('../package.json')
 
-const files = ['utilities', 'reset'];
+const env = process.env.NODE_ENV;
 
-files.forEach((file) => {
+const globals = ['reset', 'utilities'];
+const components = glob.sync('./src/components/**/styles/*.scss');
+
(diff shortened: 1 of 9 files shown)


lancetharper (approved):

lancetharper:
Do doc updates for this go in "getting started for developers"?

cowills:
@lancetharper yeah, i will update the docs when i do the "decompose utilities" ticket.

PR #709, opened 2019-10-31
Add PR template and some unit test guidelines

lancetharper:
I'm not sure where we want unit test guidelines to live, but they are included alongside the template for now.

I tried to keep them (checklist and unit test guidelines) to bare minimum so it doesn't get overwhelming and hopefully stays more in the column of 'useful' than 'annoyance'. I didn't include checks for linting, or that it passes unit tests, etc. since those will be caught by travis automatically.

There is also [this](https://github.com/stilliard/github-task-list-completed) if we want to enforce that the boxes are checked off before allowing merge that we can consider adding.

### DIFF
--- a/.github/UNIT-TEST-GUIDELINES.md
+++ b/.github/UNIT-TEST-GUIDELINES.md
@@ -6,7 +6,7 @@
 - Both positive and negative
 - Improper or inconsistent typing
   - All accepted prop types
-- Check for side effects, limit values and None
+- Check for side effects, limit values, and None
   - Really long values
   - Really short values
   - Floats and integers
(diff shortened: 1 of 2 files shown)


mhewson:
that required checklist looks interesting
would it be something that could be added and required to merge to a specific branch like next?

lancetharper:
It would apply to all PRs with a task list. Making it required to merge depends on the branch setting -- we already do it on `next` with requiring reviewers, etc. so I believe that wouldn't require us to do anything else.

We can make a separate template for PRs to master and have a different checklist too if we wanted

mhewson (approved):

mhewson (approved):

mhewson (approved):

mhewson (approved):

cowills (approved):

PR #710, opened 2019-11-05
pull reset/utils into kitchen sink

cowills:
https://rei.github.io/rei-cedar/#/breadcrumbs kitchen-sink on gh-pages works except it's missing the CSS reset so everything is **B I G**

previously the kitchen sink was depending on there already being a copy of `cedar.css` built in dist for it to use for the reset/utils. kitchen sink uses css-in-js, so it doesn't make sense to pull in the full cedar twice.

### DIFF
--- a/build/rollup-plugins.js
+++ b/build/rollup-plugins.js
@@ -33,10 +33,6 @@ if (env !== 'test') {
   copyTargets = ['static/cdr-fonts.css'];
 }
 
-if (env === 'dev') {
-  copyTargets.push('dist/cedar.css')
-}
-
 const plugins = [
   (env == 'test' || env == 'dev') && alias({
     resolve: ['.json', '.js', '.jsx', '.scss', '.vue'],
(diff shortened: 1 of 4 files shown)


lancetharper:
I think I bumped into this while working on pagination, copying the dist/cedar.css grabbed all of the component css too so when I would make design updates in the component scss it would sometimes be overridden by having the pagination css present twice as a result.

 I solved it by fixing the import in dev.js:
```
import cedarcss from './css/main.scss'; // so postcss-plugin picks it up
...
data() {
    return {
      routes,
      cedarcss,
    };
  },
```

Removing `cedar.css` from `public/index.html` (same as you)

And removing the whole `if dev === env` block from `build/rollup-plugins.js`

I'm not sure whether this way or the way I did it makes a difference though because I'm just processing `main.scss` while this is grabbing utilities/resets separately (which is all main is anyway I believe). The main difference afaict is that mine injects it into head rather than compiling it so I believe we should get hot reloading of anything from `main.scss`

lancetharper (approved):

PR #711, opened 2019-11-05
Fix/pagination

lancetharper:
## Description
Fix Pagination

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:


lancetharper:
Docs updates in this PR https://github.com/rei/rei-cedar-docs/pull/486

I want to wait until we can pull these updates in before merging that so I can verify the examples there are working properly.

src/components/pagination/CdrPagination.jsx
```diff
@@ -49,11 +61,23 @@ export default {
   },
   data() {
     return {
-      currentUrl: '',
       style,
+      componentID: Math.random().toString(36).substr(2, 9),
```
cowills, line 65: we have a bunch of repeated "unique ID" logic strewn throughout cedar, would be great to move that to 1 utility function rather than duplicating (though i don't think that needs to happen in this PR)

src/components/pagination/CdrPagination.jsx
```diff
@@ -131,30 +155,34 @@ export default {
...
     prevElAttrs() {
+      const prevPageData = this.pages[this.prevPageIdx];
       return {
-        class: clsx(this.style['cdr-pagination__link'], this.style['cdr-pagination__prev']),
-        href: this.pages[this.prevPageIdx].url,
-        page: this.pages[this.prevPageIdx].page,
-        'aria-label': 'Go to previous Page',
-        ref: 'prev-link',
+        // things that we want to be able to easily bulk bind to scoped slot (for a11y, styling, etc.)
+        attrs: {
```
cowills, line 161: very nice! 💯

src/components/pagination/CdrPagination.jsx
cowills: actually it looks like input/select are now using `this._uid` instead of generating an ID, is that something we can do here as well?

cowills: oh there's also a `guid` method inside pagination still as well. I don't have any strong opinions on whether we move that method into a mixin or just start using _uid. maybe a good critique topic for tomorrow

lancetharper: Yeah, I had to add a bunch of unique ID stuff. Some because of unit tests, some for other reasons.

I believe that using `_uid` is discouraged by the vue team since it's internal vue API.

We probably should create a guid utility because it's something I can see us using in other components in the future. I can do that here if we all agree

cowills: i think it makes sense to merge this and refactor the UID stuff into a utility later.

cowills (approved):

PR #719, opened 2019-11-20
Font updates

mhewson:
## Description
This updates tokens, type utilities, cdr-font css, and the accordion, button, link, list, input quote, tabs, pagination, cta, caption, radio, and checkbox components 

### Design:

### Cross-browser testing:
- [X] Chrome
- [X] Firefox
- [X] Edge
- [X] Safari
- [X] IE11
- [X] iOS
- [X] Android

- [X] Added/updated backstop tests
- [X] Passes with existing reference images (or failed as expected)

### Unit testing:
- [X] Sufficient unit test coverage (see unit test best practices for ideas)

### A11y:
- [X] Meets WCAG AA standards

### Documentation:


src/components/tabs/styles/vars/CdrTabs.vars.scss
```diff
@@ -0,0 +1,19 @@
+$cdr-tab-panel-animation-duration: $cdr-duration-5-x;
+
+@mixin cdr-tabs-base-header-item-label() {
+  font-family: Graphik, Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif;
```
cowills: should these values be tokens?

src/css/generic/_typography.scss
```diff
@@ -22,10 +22,97 @@ h3.cdr-text,
 h4.cdr-text,
 h5.cdr-text,
 h6.cdr-text {
+
   @include cdr-text-default-mixin;
```
cowills, line 26: should this be body-300?

src/css/index.scss
```diff
@@ -0,0 +1,18 @@
+/* Tokens */
```
cowills: is this file different from main.scss?

cowills:
are there breadcrumb updates? 
and can we delete all the redwood/spruce/etc. stuff from the _options file now?

src/components/tabs/styles/vars/CdrTabs.vars.scss
mhewson: no, these are one off values

src/css/generic/_typography.scss
mhewson: no the default is not utility-300 rather than body-300

mhewson: I'll be pushing an update to tokens to remove the deprecation and remap the default-text token

src/components/tabs/styles/vars/CdrTabs.vars.scss
cowills: do we have a token for the font-family at least?

mhewson: not at the time, all the constructors are private options and are not available to cedar -- TODO

src/css/index.scss
mhewson: this is interesting -- not sure how it got introduced -- looks like it can be removed

mhewson:
Breadcrumb, data table select and rating will be in a separate branch for ease of review, we can/ should be removing all instances of redwood and such with a resolution to this work.

src/components/tabs/styles/vars/CdrTabs.vars.scss
lancetharper: We do! It doesn't have the fallbacks, but that's probably ok to be hardcoded https://github.com/rei/rei-cedar-tokens/blob/next/dist/scss/cdr-tokens.scss#L830-L831

I added them for use in our fonts file. https://github.com/rei/rei-cedar/blob/next/src/css/fonts.scss#L65

mhewson: yes I see it now - though it would be more useful in this instance to have the entire stack tokenized rather than constructing it in this way ` font-family: $cdr-font-family-sans, Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif; `at the lease we could add the  ` cdr-font-familiy-sans-fallbacks ` token or something

mhewson: Ok I think I have resolved this for now with the latest updates

src/css/index.scss
mhewson: should be resolved

src/components/cta/styles/vars/CdrCta.vars.scss
```diff
@@ -1,5 +1,12 @@
...
+
 @mixin cdr-cta-base-text-mixin() {
-  @include maple-utility-40;
+  font-family: $cdr-cta-font-family;
+  font-style: normal;
+  font-weight: 500;
+  letter-spacing: -.08px;
+  font-size: 16px;
+  line-height: 22px;
 }
```
lancetharper, line 10: CTA isn't in our text system? 😞

src/components/tabs/styles/vars/CdrTabs.vars.scss
cowills: @mhewson @lancetharper won't we need to have a token for the full font-family stack available? what will non-component consumers do? 
if for some reason we aren't going to publish a token for this font stack, we should just create a private variable inside cedar for that so it isn't duplicated everywhere.

mhewson: Yea I will resolve this in a next PR with some updates to default text token and such

lancetharper (approved):
Adding that full stack token should be easy since we already have it defined in an option, just need to expose it here: 

option -- https://github.com/rei/rei-cedar-tokens/blob/next/tokens/_options/font.json5#L10-L12
token -- https://github.com/rei/rei-cedar-tokens/blob/next/tokens/web/font.json5

cowills (approved):

PR #725, opened 2019-11-26
delete un-used files

cowills:
- styles/components.css - ??? might have been related to the pre-single package dev build
- build/build.js - ancient webpack build script
- build/component-template-extract - ??? not being used anywhere
- build/noop.css - accidentally committed this yesterday 😆 
- src/index.html - public/index.html is the real dev server file, this one gets me every time 😠


KrisKnabel:
Can you elaborate on the `src/index.html`?  How is it getting there when the local build should be putting it in the public folder?  Is there something the team should be looking out for to not accidentally check it in?

cowills:
@KrisKnabel its an old file, idk what else to say? public/index.html is the one that the dev server uses, i have no idea where src/index.html is from

KrisKnabel:
> @KrisKnabel its an old file, idk what else to say? public/index.html is the one that the dev server uses, i have no idea where src/index.html is from

@cowills ah ok, when you said, "this one gets me every time" it sounded like it was something we needed to be removing every build or something.

KrisKnabel (approved):

PR #726, opened 2019-11-26
decompose utilities

cowills:
## Description

- Updates CSS build to generate an individual file for every category of utilities. Also creates a `dist/style/utilities.css` file which imports all of the utilities.
- renames `cedar.css` and `utilities.css` to be `cedar-compiled.css` and `utilities-compiled.css` respectively. (importing these files should be discouraged in favor of the pcss-imports style)
- renames all the utility CSS files to match the naming prefix in their tokens (i.e, typography => text, alignment => align, visibility => display)
- deletes the css/generic folder, moves it's contents into css/utility and css/reset.scss. (NOTE: not sure if there is a difference between what was in "generic" and what was in "utility"...but what we are calling "utilities" is (generic + utility - reset) so i think this makes sense?)
- update CdrText.css to pull in the typography utility file
- updates `dist/style/cdr-full.css` to import the individual utilities. 

### Design:
- n/a Reviewed with designer and meets expectations

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

- n/a Added/updated backstop tests

### Unit testing:
- n/a Sufficient unit test coverage (see unit test best practices for ideas)
- [] Manually verified in climbers-site

### A11y:
- n/a Meets WCAG AA standards

### Documentation:

WIP docs are here: https://github.com/rei/rei-cedar-docs/pull/500
waiting to do some integration testing with the beta before finalizing the update steps


mhewson (dismissed):

lancetharper (approved):
The 'generic', 'utility', etc. is from when we started out following ITCSS. We haven't really been following it since we moved component css out of that directory. All that to say that changing those shouldn't be a big deal as there was never a "mission critical" reason for choosing that directory structure.

Renaming `cedar.css` feels pretty big, but I think the reasoning makes sense especially if we're going to promote using postcss import instead

PR #728, opened 2019-11-27
Breadcrumb bugs

cowills:
## Description

- on ellipsis click, updates focus to the first breadcrumb item
- adds watcher to `items` that re-evaluates the truncation logic

docs PR: https://github.com/rei/rei-cedar-docs/pull/499

### Design:
- n/a Reviewed with designer and meets expectations

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

- n/a Added/updated backstop tests

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/src/components/breadcrumb/CdrBreadcrumb.jsx
+++ b/src/components/breadcrumb/CdrBreadcrumb.jsx
@@ -94,12 +94,15 @@ export default {
         >
           /
         </span>) : '';
+
         const attrs = {
           class: this.style['cdr-breadcrumb__link'],
         };
         if (index === this.items.length - 1) {
           attrs['aria-current'] = 'page';
         }
+
+        const ref = index === 0 ? 'firstBreadcrumb' : null;
 
         return (<li
           class={this.style['cdr-breadcrumb__item']}
@@ -111,9 +114,11 @@ export default {
               attrs,
               href: breadcrumb.item.url,
               content: breadcrumb.item.name,
+              ref,
             })
             : (<a
               {... { attrs: attrs }}
+              ref={ref}
               href={breadcrumb.item.url}
             >
               { breadcrumb.item.name }
@@ -127,6 +132,7 @@ export default {
   methods: {
     handleEllipsisClick() {
       this.truncate = false;
+      this.$nextTick(() => { this.$refs.firstBreadcrumb.focus() });
     },
   },
   render() {
--- a/src/components/breadcrumb/CdrBreadcrumb.jsx
+++ b/src/components/breadcrumb/CdrBreadcrumb.jsx
@@ -46,11 +46,6 @@ export default {
       style,
     };
   },
-  watch: {
-    items() {
-      this.truncate = this.truncationEnabled && this.items.length > 2;
-    },
-  },
   computed: {
     baseClass() {
       return 'cdr-breadcrumb';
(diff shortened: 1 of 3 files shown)


lancetharper (approved):
👍

src/components/breadcrumb/__tests__/CdrBreadcrumb.spec.js
```diff
@@ -115,4 +131,37 @@ describe('CdrBreadcrumb', () => {
...
+    ];
+    const wrapper = mount(CdrBreadcrumb, {
+      propsData: {
+        items: items,
+      },
+      attachToDocument: true, // enables focus testing
+    });
+    wrapper.vm.handleEllipsisClick()
+    wrapper.vm.$nextTick(() => {
+      expect(wrapper.vm.$refs.firstBreadcrumb).toBe(document.activeElement);
```
cowills, line 163: @lancetharper @darintrimillos @KrisKnabel  figured out a way to do focus tests in vue-test-utils, just need to use `mount` instead of 'shallowMount', set `attachToDocument`, and then `document.activeElement` will be whatever currently has focus (might also need to use `async` and `nextTick` depending on what the code is doing)

src/components/breadcrumb/CdrBreadcrumb.jsx
```diff
@@ -42,17 +42,14 @@ export default {
   },
   data() {
     return {
-      shouldTruncate: this.truncationEnabled && this.items.length > 2,
+      truncate: this.truncationEnabled && this.items.length > 2,
```
KrisKnabel, line 45: I think this is the reason why the SSR version of the breadcrumb was not truncating.  This is not reactive when in the data object.  We should move this to a computed property that checks `this.truncationEnabled && this.items.length > 2` and I think that will fix the defect product was talking about.

src/components/breadcrumb/__tests__/CdrBreadcrumb.spec.js
KrisKnabel: Awesome!

src/components/breadcrumb/CdrBreadcrumb.jsx
```diff
@@ -116,9 +117,15 @@ export default {
       });
     },
   },
+  watch: {
+    items() {
+      this.truncate = this.truncationEnabled && this.items.length > 2;
+    },
+  },
```
KrisKnabel, line 124: Ah or this works too I guess, not exactly sure if one solution is better than the other but if this works I'm ok with going this route.

KrisKnabel: I've read that computed props do have some extra cacheing logic under the hood, but with this being such a simple computation we likely won't see any benefit.

KrisKnabel (approved):
Left some comments, but I think this is good to merge at your discretion.

src/components/breadcrumb/CdrBreadcrumb.jsx
cowills: we need to have a data prop for the `onEllipsisClick` method to work, since the component has internal state.

KrisKnabel: Yes, but tying part of that data property value to the value of the computed prop would get rid of the need of the watcher.  There are also computed prop setters that could be used: https://vuejs.org/v2/guide/computed.html#Computed-Setter.  I was just mostly bringing this up as I had recently referenced the docs and it appears Vue is pretty opinionated about using Computed Props over Watchers when possible.

PR #730, opened 2019-12-05
Add size to checkbox and radio

lancetharper:
## Description

Checkbox and Radio now accept a size prop.

Other questions:
- Do we want to deprecate 'compact' (in favor of small)?
- I added mixins for the size variants that should make using them via component variables easier but they might be a bit too clever

I'll update radio/checkbox API to show they now support size once this is merged

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/src/components/checkbox/styles/vars/CdrCheckbox.vars.scss
+++ b/src/components/checkbox/styles/vars/CdrCheckbox.vars.scss
@@ -1,11 +1,11 @@
-@mixin cdr-checkbox-text {
+@mixin checkbox-text {
   @include cdr-text-utility-300;
   color: $cdr-color-text-form-label-lightmode;
   display: inline-block;
   margin: 0;
 }
 
-@mixin cdr-checkbox-text-compact {
+@mixin checkbox-text-compact {
   @include cdr-text-utility-200;
 }
 
@@ -35,7 +35,7 @@ $checkbox-compact-figure-space-top: 2px;
 $checkbox-figure-checked-hover-background-color: #2b6692;
 $checkbox-focus-outline: 0 0 8px 0 rgb(50, 120, 174);
 
-@mixin cdr-checkbox-small-mixin($parent-selector, $breakpoint: '') {
+@mixin checkbox-small-mixin($parent-selector, $breakpoint: '') {
   &--small#{$breakpoint} {
     @include cdr-text-utility-200;
     padding-left: calc(#{$checkbox-figure-size-small} + #{$cdr-label-base-spacing});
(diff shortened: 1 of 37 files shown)


src/components/checkbox/styles/vars/CdrCheckbox.vars.scss
```diff
@@ -1,11 +1,24 @@
...
+  color: $cdr-color-text-form-label-lightmode;
+  display: inline-block;
+  margin: 0;
+}
+
+@mixin checkbox-text-compact {
+  @include cdr-text-utility-200;
+}
+
+$checkbox-figure-size-small: 16px;
```
cowills, line 12: should we be using the `cdr-icon-size` tokens here?

src/components/radio/__tests__/__snapshots__/CdrRadio.spec.js.snap
```diff
@@ -5,7 +5,7 @@ exports[`CdrRadio renders an input 1`] = `
   class="cdr-radio__wrap"
 >
   <label
-    class="cdr-radio"
+    class="cdr-radio cdr-radio cdr-radio--medium"
```
cowills: duplicate class here (happening in the checkbox snapshot too)

src/components/checkbox/styles/vars/CdrCheckbox.vars.scss
lancetharper: Maybe, but since the large variant isn't the same size as a large icon I'm not sure they're intended to match that way

src/components/radio/__tests__/__snapshots__/CdrRadio.spec.js.snap
lancetharper: I think this is coming from the size mixin. Looks like the same thing happens to button which also uses it

cowills: is that a bug in size/buildClass then?

lancetharper: Yeah, buildClass wasn't intended to be used as a catch all for all mixins originally so it always added the baseClass. I'll make a different branch which fixes that everywhere and we can consider what to do there.

cowills (approved):
LGTM. I do think we should deprecate compact in favor of small, but that might require a :+1: from design. in any case, that wouldn't require any work in this release besides calling it out in the API docs/release notes.

PR #734, opened 2019-12-18
feat(tabs): add centered tab variant, show scrollbar on osx, address seo issue

lancetharper:
## Description

Add centered modifier. 

Adds CSS hack to show scrollbar when content overflows on OSX CDR-967

fixes issue with underline when header overflows

fixes issue with header scrolling on windows chrome

Updates the tab `href`s to use a hash link based on the id or name of the tab, and also adds the respective ID to it's tab content container. We use `click.prevent` on the tab links here, so currently the hash doesn't really "do" anything (i.e, it won't update the URL unless you manually listen for the tab-change event.) Setting the href to a hash should address the SEO concerns for now. Since the links aren't actually "navigational", we don't need to implement a method for setting the active tab on load (though that would be trivial to do by exposing `activeTabIndex` on tabs).

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:


src/components/tabs/CdrTabs.jsx
```diff
@@ -46,14 +45,20 @@ export default {
       .filter(tab => tab); // get vue component children in the slot
 
     this.$nextTick(() => {
-      this.initializeOffsets();
-      this.headerWidth = this.getHeaderWidth();
       if (this.tabs[0] && this.tabs[0].setActive) this.tabs[0].setActive(true);
+
+      // nextTick does not work in firefox here
+      setTimeout(() => {
```
cowills, line 54: need to do more testing on this. a setTimeout of 1 fixes the bug in firefox, around 100 is necessary for ios. 

i believe this is because the `mounted` event is firing before all of this tabs content has been mounted (possibly due to the `this.tabs` stuff going on?)

cowills: 100ms timeout works in every browser/device combo i’ve tried. since the mounted event doesn't run server side it should not impact SSR to do things like this. 

The underline bug does seem to be related to the tab header generation stuff we are doing (i suspect that `.map(vnode => vnode.componentInstance)` is causing things to happen outside the usual vue lifecycle, which is why nextTick does not suffice here...) Refactoring this component to use slots would likely solve this issue.

there is a “double RAF” technique (`requestAnimationFrame(() => { requestAnimationFrame(callback); });` ) which i found that seems to work in firefox and newer IOS but not older IOS versions/devices. this seems to confirm that this is a race condition involving content being rendered in a way that vue cannot track.

an alternative might be to run a loop that waits for the tab header elements to be fully rendered on the page before running updateUnderline, however i do not see anything we could hook into to do that (the elements still report an `offsetLeft` and `width`, they're just wrong because the element hasn't been rendered/styled yet)

darintrimillos (approved):

src/components/tabs/CdrTabs.jsx
lancetharper: Yeah, it's 100% because it's trying to manipulate child components in the slot (we shouldn't really be doing it this way) but this works for now

PR #742, opened 2020-01-02
fix(build): remove @import url statements from compiled CSS

cowills:
this is a lil hacky but im not sure there is non hacky way to solve this but i am open to suggestions for alternative hacks 😆 

- we currently are using `@import url()` to handle loading CSS for cedar components that depend on other cedar components (for example, accordion uses icon: https://github.com/rei/rei-cedar/blob/548c3a0f675b29572ecba63cfc38e17424c01a24/src/components/accordion/styles/CdrAccordion.scss#L22)
- this can cause an issue if you are trying to use the `cedar-compiled.css` in an environment that is not set up to resolve those imports. (only popped up in component variables, doesn't seem to be an issue in the sandboxes or docs)
- however, because the cedar-compiled file has all of the cedar CSS in it, we don't have to worry about component dependencies and can just ignore the @import url statements inside cedar (which is what we are functionally doing by forcing all of those imports to load an empty file)

### DIFF
--- a/build/noop.css
+++ b/build/noop.css
@@ -1,0 +1,1 @@
+/* intentionally empty */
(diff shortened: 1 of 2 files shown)


lancetharper (approved):
Nothing better immediately comes to mind. This is pretty small so if we do end up changing (or having to change) this is at least super easy to undo 👍

PR #745, opened 2020-01-08
fix(modal): ie11 close modal with Esc

darintrimillos:
fix CDR-1516

## Description
IE11 keyboard event fires 'Esc' instead of 'Escape'

### Design:

### Cross-browser testing:
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [x] IE11
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [ ] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/src/components/modal/CdrModal.jsx
+++ b/src/components/modal/CdrModal.jsx
@@ -96,8 +96,6 @@ export default {
     handleKeyDown({ key }) {
       switch (key) {
         case 'Escape':
-          this.onClick();
-          break;
         case 'Esc':
           this.onClick();
           break;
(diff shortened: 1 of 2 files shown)


src/components/modal/CdrModal.jsx
```diff
@@ -98,6 +98,9 @@ export default {
         case 'Escape':
```
lancetharper, line 98: It looks like key has good compatibility https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

You don't need a separate case for this, you can just stack them so long as they do the same thing:
```
case 'Escape':
case 'Esc':
...
break;
```

darintrimillos: Ah, ok. I'll fix that.

cowills (approved):

PR #746, opened 2020-01-08
Update tokens again

cowills:
none of the tokens being removed had been released yet, so no real changes needed anywhere


lancetharper (approved):
😓

PR #751, opened 2020-01-13
add postcss-import-able output for deprecated utility classes

cowills:
currently the deprecated utility classes are not included in the de-composed CSS output. 
adding this output will give consumers a way to manually load these deprecated classes in their project if they need to.


lancetharper (approved):
Good catch

PR #757, opened 2020-01-15
fix(caption): cite browser default is font style italic design would…

mhewson:
… like this to be normal

### Design:

### Cross-browser testing:
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE11
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [ ] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/src/components/caption/styles/CdrCaption.scss
+++ b/src/components/caption/styles/CdrCaption.scss
@@ -15,6 +15,10 @@
   max-width: 530px;
   line-height: 1;
   padding: $cdr-space-inset-one-x-squish;
+  
+  cite {
+    font-style: normal;
+  }
 
   &__cite {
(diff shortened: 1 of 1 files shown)


lancetharper:
Should this be all cites or just within the caption component?

mhewson:
I thought of placing this in reset but thought it could be a bit heavy handed as they are specifically calling out this instance in which we have used a `cite` tag

cowills (approved):

lancetharper:
This should have used the class and not styled the element tag

cowills:
> This should have used the class and not styled the element tag

its scoped to this component, so it only applies to `cite`'s inside a cdr-caption.

PR #760, opened 2020-01-15
Fix wonky ratings

lancetharper:
## Description

Ratings stars were stacked on each other (empty then filled over top) which caused some rendering issues for some people

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

- [ ] Added/updated backstop tests (N/A)

### Unit testing:
Some unit tests that are caught by backstop tests have been removed

### A11y:
- [x] Meets WCAG AA standards

### Documentation:
- [ ] API docs created/updated (N/A)

### DIFF
--- a/src/components/rating/__tests__/CdrRating.spec.js
+++ b/src/components/rating/__tests__/CdrRating.spec.js
@@ -22,21 +22,25 @@ describe('CdrRating', () => {
     expect(wrapper.vm.whole).toBe(3);
     expect(wrapper.vm.remainder).toBe('50');
     expect(wrapper.vm.rounded).toBe(3.5);
-
+    expect(wrapper.vm.empties).toBe(1);
+    
     wrapper.setProps({rating: 3.122227});
     expect(wrapper.vm.whole).toBe(3);
     expect(wrapper.vm.remainder).toBe('00');
     expect(wrapper.vm.rounded).toBe(3);
-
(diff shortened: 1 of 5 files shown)


src/components/rating/__tests__/CdrRating.spec.js
```diff
@@ -61,52 +65,6 @@ describe('CdrRating', () => {
...
   });
 
-  it('renders normal placeholder stars when count is > 0', () => {
-    const wrapper = shallowMount(CdrRating, {
-      propsData: {
-        rating: 5,
-        count: 1,
-      }
-    });
-    expect(wrapper.contains('.cdr-rating__placeholder')).toBe(true);
```
cowills, line 71: aren't these tests still valid? i think they cover the wonky display logic that rating has

lancetharper: No, the logic had to change. Before we rendered 5 `placeholder--no-reviews` (which are gray) always.

Now we only render 5 gray stars when there are 0 reviews and a 0 rating (no rating). Otherwise if rating is 3 it would have rendered 3 blue stars and 2 gray stars when what we want are all blue.

I added more examples to catch these cases so they should now be covered by backstop (which is probably better than just looking for classes anyway)

cowills (approved):

PR #762, opened 2020-01-29
fix(select): update vmodel logic to not directly modify a prop, emit a different unique event name

lancetharper:
## Description

Updates to avoid double events.

- Emit new event `select-change` (avoids possible collisions with input/change)
    - Old events are still emitted, but considered "deprecated"
- Use computed prop for internal vmodel to avoid mutating value prop
- Update model event to `update-select` to avoid possible collisions

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [ ] Meets WCAG AA standards

### Documentation:

### DIFF
--- a/src/components/select/CdrSelect.jsx
+++ b/src/components/select/CdrSelect.jsx
@@ -12,6 +12,10 @@ export default {
   },
   mixins: [size, space],
   inheritAttrs: false,
+  model: {
+    prop: 'value',
+    event: 'update-select',
+  },
   props: {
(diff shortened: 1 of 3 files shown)


lancetharper:
Will roll this into pagination instead since they are dependent and this fails those unit tests :(

PR #763, opened 2020-01-31
Pagination/Select updates

lancetharper:
## Description

I believe this fixes all of the issues. But please pull this and double check me 😄 

Select:
- emits new `select-change` event to avoid collisions **needs doc update**
- updated to not mutate the prop directly
- custom model event to avoid possible collisions (`update-select`)
- cleaned up input event binding (not needed)
- emits old events (`change`, `input`) for backwards compatibility **needs doc update to mark as deprecated**
- updated tests

Pagination:
- updated to use new select event
- removed `requestAnimationFrame` code (culprit for blocking preventDefault)
- updated vue-router example to not use scopedSlot/`router-link` and use programatic navigation instead **needs doc update for recommended method**
  - this can probably be done another way, but this works while `router-link` was causing race conditions with with our click event and timing when the route updated
- updated logic for tracking current internal state
- updated tests

I created this sandbox to test against Brian's use case -- it's using the built file from this code.
https://codesandbox.io/s/pagination-sandbox-z6o0p

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:


src/components/select/CdrSelect.jsx
```diff
@@ -94,38 +106,26 @@ export default {
         {},
         this.$listeners,
         {
-          input(event) {
+          change(event) {
```
cowills, line 101: https://github.com/rei/rei-cedar/blob/next/src/components/input/CdrInput.jsx#L89-L102 Do we know why the `input` event works for `input` but `change` doesn't work here?

lancetharper: I think it's because input only reacts to the input event. Select was reacting to both input and change (maybe because of v-model, I'm not really sure)

We don't even call this out as being emitted by input in the docs anymore, we should just remove it and pass all the listeners because consumers can't get the actual browser input event right now, just the value. 

If we really don't care about `multiple`, we could also get rid of it for select too since we really only need it for doing the work of getting those values for consumers.

cowills: but if we made select only respond to one of the events, could it work the same as input does? (we should also probably be passing the event in input so consumers can access it)

lancetharper: I'll add a card for dev critique and we can go over all of this next week 👍

lancetharper:
These latest changes have those updates to select @cowills. One thing to note is that our change event is different than the regular vue change event because ours emits the value & event while the regular vue one is just the event. We'll need to at least add docs for change because it works differently than normal because of that.

cowills (dismissed):
this LGTM. I searched bitbucket for `"cdr-select" ext:vue` and couldn't find any examples of cdr-select usage that would break with this change, since everyone is either listening for `@change` or just letting the v-model do all the work. nobody was touching @input (though they could still listen for that event if they wanted to. 

next major release we might wanna consider switching to the `('eventName', event, value)` event signature so we can leverage more of the vue magic.

should we cut a beta and verify it resolves the issues on content-hub?

lancetharper:
Yeah, I can publish one. 4.0.2? Or are we calling it 4.1.0?

cowills:
> Yeah, I can publish one. 4.0.2? Or are we calling it 4.1.0?

i think 4.0.2 works since its a bug fix

lancetharper:
4.0.2-beta.0 is up

cowills (approved):

PR #764, opened 2020-01-31
fix(type): add breakpoint suffixes to responsive heading-serif-strong

cowills:
Bug pointed out by Jake Rohr in user support
For the patch release


lancetharper (approved):
We should really make this function/mixin driven in the future

PR #765, opened 2020-01-31
de-variable-ize

cowills:
## Description
- Deleting overly specific component variables and hardcoding their values instead
- Identifying which component variables could possibly be replaced with either a generic or specific token. 

### Component Re-mapped

- [x] Accordion
- [x] Breadcrumb
- [x] Buttons
- [x] Call to Action
- [x] Caption
- [x] Checkboxes
- n/a Data Tables
- n/a Grid
- n/a Icons
- n/a Images
- [x] Inputs
- [x] Links
- [x] Lists
- [x] Modal
- n/a Pagination
- [x] Quote
- [x] Radio Buttons
- [x] Ratings
- [x] Select
- [x] Tabs


src/components/radio/styles/CdrRadio.scss
```diff
@@ -49,24 +49,23 @@
   ========== */
 
   &__figure {
+    @include cdr-label-figure-medium-mixin
     position: absolute;
-    top: $radio-figure-space-top;
```
cowills, line 53: $radio-figure-space-top is 4px https://github.com/rei/rei-cedar/blob/next/src/components/radio/styles/vars/CdrRadio.vars.scss#L21

However in the medium size mixin (which is also the "default" or "base" size), top is set to `calc((#{$cdr-text-utility-300-height} - #{$radio-figure-size-medium}) / 2);` which is `(2.2rem - 16px) / 2` => 3px
https://github.com/rei/rei-cedar/blob/next/src/components/radio/styles/vars/CdrRadio.vars.scss#L54

This pr updates it so that the default and medium styles are the same

cowills:
@lancetharper @darintrimillos this shouldn't get merged till after the patch but it is ready to start reviewing. It refactors a LOT of the SCSS so we will want to merge this ASAP before any styling related work for spring release begins.

cowills:
@lancetharper @darintrimillos @mhewson This should be ready to review and merge now that next is clear of the patch. 

We will definitely want to merge this before starting any big color work or other work involving styles. This PR is broken up so each commit corresponds to one component, so we also have the option of cherry-pick'ing these changes into next as we do the color update work if we'd prefer to do merge things incrementally.

lancetharper (approved):
I've looked through it the last couple days and I vote we just merge it to have everything done/ready

PR #767, opened 2020-02-05
refactor(select): make select component use change event

cowills:
poking around on next, this seems like it stops select from emitting duplicate events, but im not sure of all the cases to test. i suspect that the native select using change and our select using input might have been causing weird behavior. (i.e, the native select would emit an `input` and a `change` and then the opposite listeners would re-emit another one)

### DIFF
--- a/src/components/select/CdrSelect.jsx
+++ b/src/components/select/CdrSelect.jsx
@@ -12,6 +12,10 @@ export default {
   },
   mixins: [size, space],
   inheritAttrs: false,
+  model: {
+    prop: 'value',
+    event: 'change',
+  },
   props: {
     /**
      * `id` for the select that is mapped to the label `for` attribute. If one is not provided, it will be generated.
@@ -94,37 +98,9 @@ export default {
         {},
         this.$listeners,
         {
-          input(event) {
-            if (vm.multiple) {
-              const optArr = toArray(event.target.options);
-              const selected = optArr.filter(o => o.selected === true).map(o => o.value);
-              vm.$emit('input', selected, event);
-
-              // Deprecated Event
-              vm.$emit('change', selected, event);
-
-              vm.value = selected;
-            } else {
-              vm.$emit('input', event.target.value, event);
-
-              // Deprecated Event
-              vm.$emit('change', event.target.value, event);
-            }
-          },
           change(event) {
             // Deprecated event
             vm.$emit('change', event.target.value, event);
-
-            // Needed for Internet Explorer
-            if (vm.value !== event.target.value) {
-              if (vm.multiple) {
-                const optArr = toArray(event.target.options);
-                const selected = optArr.filter(o => o.selected === true).map(o => o.value);
-                vm.$emit('input', selected, event);
-              } else {
-                vm.$emit('input', event.target.value, event);
-              }
-            }
           },
         },
       );


lancetharper:
The issue with this is that we were switching from `@change` to `@input`...but it looks like we don't even say that anymore; we just say "all listeners are passed through". I can try this out on my branch if we don't care that it _may_ be "breaking"

cowills:
> The issue with this is that we were switching from `@change` to `@input`...but it looks like we don't even say that anymore; we just say "all listeners are passed through". I can try this out on my branch if we don't care that it _may_ be "breaking"

I'd rather break consumers than create a custom event type just for select. AFAICT the breaking change would potentially just be changing `@input` to `@change` and possibly updating the function signature which doesn't seem too bad.

src/components/select/CdrSelect.jsx
```diff
@@ -12,6 +12,10 @@ export default {
   },
   mixins: [size, space],
   inheritAttrs: false,
+  model: {
+    prop: 'value',
+    event: 'change',
```
lancetharper, line 17: I think this should stay unique -- no one listens to this event and it's just for keeping the internal/external value in sync (so we aren't mutating the value prop directly)

src/components/select/CdrSelect.jsx
```diff
@@ -94,37 +98,9 @@ export default {
...
-            } else {
-              vm.$emit('input', event.target.value, event);
-
-              // Deprecated Event
-              vm.$emit('change', event.target.value, event);
-            }
-          },
           change(event) {
             // Deprecated event
             vm.$emit('change', event.target.value, event);
```
lancetharper, line 103: So long as we don't care about emitting other events, then just having this one works fine. I think we should move the event to be the first arg though so then it actually works similarly to how the default one does. Then people can do the `@change.prevent` stuff

lancetharper:
I have this all working on the `pagination-fix` branch and I can push these changes there if you want to see them in context

src/components/select/CdrSelect.jsx
cowills: i think this just needs to match whatever event the native select and inputListeners is targeting. it could be `input`, `select-change`, etc.

lancetharper: That's only if we are mutating the value prop directly (which the linter says we shouldn't) and not doing it via a computed prop or watch like this: https://github.com/rei/rei-cedar/blob/pagination-fix/src/components/select/CdrSelect.jsx#L62-L69

src/components/select/CdrSelect.jsx
cowills: that makes sense to me, as long as the v-model magic still works (though we should update input/radio/checkbox to match that event signature too)

src/components/select/CdrSelect.jsx
cowills: this is what the default v-models settings are:
```
model: {
prop: 'value',
event: 'input'
}
```
https://github.com/rei/rei-cedar/blob/next/src/components/input/CdrInput.jsx it's working rn in input, im just confused why it doesn't work in select.

lancetharper: It does work in select, I think it's just not recommended to mutate props directly. I think it's fine for these input components though because we want to keep them kind of transparent. However, for other things like pagination that also use v-model we should control when/how the internal and external values are updated and work.

PR #769, opened 2020-02-08
fix(line-height): this ensures the text is dead center on the icons w…

mhewson:
…ithin rating, this update will

### Design:

### Cross-browser testing:
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE11
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [ ] Meets WCAG AA standards

### Documentation:


lancetharper:
This "works" for the current typography because it sets the line-height to be greater than the height of the icons. If/when typography changes or icon size changes this may not work anymore.

If we're ok with having magic numbers like this, then I guess we just accept that...but this is still just a band-aid and doesn't fix the underlying issue

darintrimillos:
> This "works" for the current typography because it sets the line-height to be greater than the height of the icons. If/when typography changes or icon size changes this may not work anymore.
> 
> If we're ok with having magic numbers like this, then I guess we just accept that...but this is still just a band-aid and doesn't fix the underlying issue

I'm OK with the band-aid because it seems like it might be a while before we have a better fix figured out.

src/components/rating/styles/vars/CdrRating.vars.scss
```diff
@@ -8,12 +8,15 @@ $rating-number-decoration: $cdr-color-border-secondary-lightmode;
 
 @mixin cdr-rating-base-text() {
   @include cdr-text-utility-200;
+  line-height: 1.8;
```
lancetharper, line 11: This is the only change that actually affects the height afaict. The others just set the line-height equal to the icon height which doesn't change anything alignment-wise that I can see

mhewson: true for the fonts demo'ed but having the heights of each exactly equal seems like the best way to ensure that nothing ever moves them out of alignment

cowills:
@mhewson is this ready to review and merge?

mhewson:
@cowills,  I'm planning on going over this s at review -- fixing this has other impacts to alignment with other text that I want design to be aware of

cowills:
@mhewson can we handle the a11y fixes in a separate PR? i don't want that to block the patch release

PR #778, opened 2020-02-13
CdrTabs updates

darintrimillos:
## Description
This PR is for CDR-1604 and CDR-1603. There are followup stories for accessibility, as well as some other design updates.

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [ ] Meets WCAG AA standards

### Documentation:


src/components/tabs/styles/CdrTabs.scss
```diff
@@ -133,6 +133,17 @@
       text-decoration: none;
       color: $cdr-color-text-link-lightmode;
     }
+
+    &--disabled {
+      color: #D1CBBD;
```
darintrimillos, line 138: This slipped through... I'll correct it in the next batch of tab work next sprint.

src/components/tabs/CdrTabs.jsx
```diff
@@ -2,16 +2,21 @@ import debounce from 'lodash-es/debounce';
...
 export default {
   name: 'CdrTabs',
-  mixins: [modifier],
+  mixins: [modifier, size],
   props: {
     height: {
       type: String,
       default: '240px',
     },
+    activateTab: {
```
cowills, line 16: if we make this default to 0 then you wouldn't have to check whether or not it's passed, you can just do `activeTabIndex = this.getNextTab(activateTab)`

`activeTab` or smth might work better. `initialActiveTabIndex` would be most semantic but feels bad.



im not saying this is a "good" idea but...make it an optional v-model with a default value of 0? (then consumers could programmatically change the active tab after load as well, which...idk if that is good or bad)

cowills:
do the tabs no longer do the thing where they're aware of which direction they're sliding in from? wonder about the a11y implications of doing animations like that, ifl if we have any kind of motion it should match the direction things are moving in.

darintrimillos:
> 
> 
> do the tabs no longer do the thing where they're aware of which direction they're sliding in from? wonder about the a11y implications of doing animations like that, ifl if we have any kind of motion it should match the direction things are moving in.

that's right, they don't do the same motion thing anymore. i think this is a good point; these values are probably not the final values as it's still going to go through some final design review. i'll bring this up with design.

cowills:
> that's right, they don't do the same motion thing anymore. i think this is a good point; these values are probably not the final values as it's still going to go through some final design review. i'll bring this up with design.

is it possible to split up the animation updates from the other changes? i don't want to hold up the disabled/activeTab/size/a11y stuff

darintrimillos:
Opened a different PR without the animation changes.

PR #781, opened 2020-02-21
Breadcrumb a11y

lancetharper:
## Description

This fixes the issues listed in CDR-1618 and adds unit tests for aria attributes and such.

I still think it could be better; like I'm unsure what to do with a button that just disappears when clicked -- do we need aria-live or something with the other links just poofing into existence?

@mhewson Maybe we need a second pass with these updates in place to see if there is anything else needed?

### Design:
- [ ] ~Reviewed with designer and meets expectations~

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

- [ ] ~Added/updated backstop tests~

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:
- [ ] ~API docs created/updated~
- [ ] ~Examples created/updated~

### DIFF
--- a/src/components/breadcrumb/styles/CdrBreadcrumb.scss
+++ b/src/components/breadcrumb/styles/CdrBreadcrumb.scss
@@ -19,6 +19,7 @@
 ========================================================================== */
 @import '../../../css/settings/index.scss';
 @import './vars/CdrBreadcrumb.vars.scss';
+@import url('@rei/cedar/dist/style/cdr-icon.css');
 
 .cdr-breadcrumb {
   line-height: 1;
--- a/src/components/breadcrumb/styles/CdrBreadcrumb.scss
+++ b/src/components/breadcrumb/styles/CdrBreadcrumb.scss
@@ -20,6 +20,7 @@
 @import '../../../css/settings/index.scss';
 @import './vars/CdrBreadcrumb.vars.scss';
 @import url('@rei/cedar/dist/style/cdr-icon.css');
+@import url("@rei/cedar/dist/style/display.css");
 
 .cdr-breadcrumb {
   line-height: 1;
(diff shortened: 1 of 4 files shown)


cowills:
`aria-live="polite"` might be what we want on any hidden links?

what if we made it so the ellipsis toggle is visually hidden but still present for screenreaders when it is clicked? https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element and then just set aria-expanded on the ellipsis depending on the state of the breadcrumb? would also need to make it so that clicking the ellipsis a second time can "re-truncate" things for screenreader users

lancetharper:
I took some cues from this: https://accessabilly.com/breadcrumb-trails-come-in-and-find-out/

Basically, I think our implementation of showing/hiding is ok (it works basically the same as that one anyway) and added a couple more aria tags and some more helpful text.

That article also mentions having the last thing in the list NOT be a link to the same page that it's on for SEO purposes so I made that update as well.

src/components/breadcrumb/styles/CdrBreadcrumb.scss
```diff
@@ -20,6 +20,7 @@
 @import '../../../css/settings/index.scss';
 @import './vars/CdrBreadcrumb.vars.scss';
 @import url('@rei/cedar/dist/style/cdr-icon.css');
+@import url("@rei/cedar/dist/style/display.css");
```
cowills: we shouldn't use utility classes in the components if we can avoid it. (im not necessarily opposed to using utils in the components, but i think we should either use them everywhere, or not use them at all)

https://github.com/rei/rei-cedar/blob/next/src/components/rating/styles/CdrRating.scss#L57-L59 is an example of how we've handled doing screenreader only stuff before.

cowills: a11y improvements LGTM otherwise :+1: to this pr

lancetharper: That actually wasn't being used anymore, I just missed cleaning it up from testing. Removed it

cowills (dismissed):

cowills (approved):

PR #782, opened 2020-03-02
tabs updates without animation change

darintrimillos:
## Description
This branch has all the tabs changes, except for the animation changes, which will get further reviewed by design. I also updated based on @cowills feedback to rename activateTab and simplify the setup in the mounted hook.

Jira Tickets:
CDR-1616
CDR-1617
CDR-1603
CDR-1604

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [ ] iOS
- [ ] Android

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:


src/components/tabs/CdrTabs.jsx
```diff
@@ -44,6 +49,8 @@ export default {
       .map(vnode => vnode.componentInstance)
       .filter(tab => tab); // get vue component children in the slot
 
+    this.activeTabIndex = this.getNextTab();
```
cowills: `this.activeTabIndex = this.getNextTab(this.activeTab);` for activeTab to work

darintrimillos: Whoops. Fixed.

lancetharper:
I know this is meant to be links, but I wonder if we need to emit a click event too like pagination for people using vue-router and other routing oddities

cowills:
> I know this is meant to be links, but I wonder if we need to emit a click event too like pagination for people using vue-router and other routing oddities

i think we could solve that in a generic way by giving full programmatic control over which tab is active...though i can't really think of a case where someone would want or need to do that.

lancetharper:
> > I know this is meant to be links, but I wonder if we need to emit a click event too like pagination for people using vue-router and other routing oddities
> 
> i think we could solve that in a generic way by giving full programmatic control over which tab is active...though i can't really think of a case where someone would want or need to do that.

Wouldn't you still have to `preventDefault` the link click though?

cowills:
> Wouldn't you still have to `preventDefault` the link click though?

yeah, i guess my point is if we want to support that there's other work to do besides exposing the event, and i can't think of a use case for it either

cowills (dismissed):

cowills (approved):

PR #789, opened 2020-03-05
Update color for text-ish components

cowills:
## Description

These components seemed like the lowest hanging fruit and least likely to change.

Link, List, Breadcrumb, Caption, Quote, CSS Reset

Open Questions:
- we currently do not set color: primary in caption, list, or quote, and just allow them to inherit their color. Should we continue that or explicitly set color: primary on those? **answer: set caption and quote to primary, let list inherit**
- i just deleted the `on-light` and `on-dark` options for cdr-link. I couldn't find any instance of those being used with link in bitbucket, and we don't have it documented anywhere. (there was a single instance of someone using cdr-button on-dark but otherwise i couldn't find any one using those theming classes). **answer: will discuss at dev critique, note in release notes just in case**
- we have designs and tokens for link disabled, but no way to handle it currently unless we break the API or add a new property. (we could say "a link without an href is disabled" but ppl use links for JS triggers, cdr-link href defaults to "#", and tag="button" results in no href being set) **Opening a separate ticket for adding a disabled link feature.**
- i was able to update the unordered list elements to use the secondary color, we can do the same to the ordered numbers as well if we want to, would just require a lil CSS: https://css-tricks.com/custom-list-number-styling/) **answer: Opening a separate PR/ticket for this, as it seems a lil tricky to get it to match the current list styles**

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

### Unit testing:
- n/a Sufficient unit test coverage (see unit test best practices for ideas)

### A11y:
- [x] Meets WCAG AA standards

### Documentation:
- n/a Examples created/updated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1392,9 +1392,9 @@
       "dev": true
     },
     "@rei/cdr-tokens": {
-      "version": "4.0.0-alpha.2",
-      "resolved": "https://registry.npmjs.org/@rei/cdr-tokens/-/cdr-tokens-4.0.0-alpha.2.tgz",
-      "integrity": "sha512-REieNA+hNnqF2XCjPbqStUa1vpd8uX1IhMTZ8SFqWp6rNl1F8Fichj+rzU/lohmBNgtj9GO066ohAkh5Y920/w==",
+      "version": "4.0.0-alpha.3",
+      "resolved": "https://registry.npmjs.org/@rei/cdr-tokens/-/cdr-tokens-4.0.0-alpha.3.tgz",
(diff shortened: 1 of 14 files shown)


mhewson (dismissed):

mhewson:
looks good so far

lancetharper:
Color updates look good, but sounds like we need a broader discussion around text color inheriting vs explicitly setting it -- I vote we use dev critique for that and hold off merging for now

lancetharper (approved):

PR #795, opened 2020-03-09
Tab panel accessibility fix

darintrimillos:
## Description
Adds support for aria-controls and aria-labelledby.

### Design:
n/a - no design changes

### Cross-browser testing:
n/a - this is just updated accessibility

- [ ] ~Passes with existing reference images (or failed as expected)~

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:
- [ ] ~API docs created/updated~

### DIFF
--- a/src/components/tabs/CdrTabPanel.jsx
+++ b/src/components/tabs/CdrTabPanel.jsx
@@ -34,7 +34,6 @@ export default {
     return {
       active: false,
       offsetX: 0,
-      tabId: this.id || this.name,
       animationDirection: 'default',
       style,
     };
@@ -56,8 +55,8 @@ export default {
   methods: {
     setActive(state) {
       this.active = state;
-      this.$emit('tab-change', state, this.tabId);
-      this.$emit('tabChange', state, this.tabId);
+      this.$emit('tab-change', state, this.id);
+      this.$emit('tabChange', state, this.id);
     },
     setAnimationDirection(direction) {
       this.animationDirection = direction;
@@ -100,7 +99,7 @@ export default {
           aria-hidden={!this.active}
           aria-labelledby={this.ariaLabelledBy}
           class={clsx(this.style[this.baseClass], this.modifierClass)}
-          id={this.tabId}
+          id={this.id}
           ref="cdrTabPanelContainer"
           tabindex="0"
           role="tabpanel"
--- a/src/components/tabs/CdrTabPanel.jsx
+++ b/src/components/tabs/CdrTabPanel.jsx
@@ -17,7 +17,7 @@ export default {
       type: Boolean,
       default: false,
     },
-    ariaLabelledBy: {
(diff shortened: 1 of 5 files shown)


src/components/tabs/CdrTabPanel.jsx
```diff
@@ -17,17 +17,23 @@ export default {
       type: Boolean,
       default: false,
     },
+    ariaLabelledBy: {
```
lancetharper: this should be `ariaLabelledby` so it matches the normal attribute -- less confusing imo

darintrimillos: Makes sense. I'll fix it.

lancetharper (approved):

src/components/tabs/examples/Tabs.vue
```diff
@@ -41,31 +48,47 @@
...
             tag="strong"
             modifier="subheading"
           >
             Tab Five Content
           </cdr-text>
         </cdr-tab-panel>
-        <cdr-tab-panel name="Six">
+        <cdr-tab-panel
+          name="Six"
+          id="tab-panel-default-6"
```
cowills: does it make sense for the id and aria-labelledby to be the same here? don't they get applied to the same element?

src/components/tabs/CdrTabPanel.jsx
```diff
@@ -17,17 +17,23 @@ export default {
...
+      required: true,
+    },
     /**
      * Optional extra reference value to be set on the tab, otherwise
      * the tab name will be used for reference.
      */
-    id: String,
+    id: {
+      type: String,
+      required: true,
```
cowills, line 30: is there a reason for making this required? what does that solve?

src/components/tabs/CdrTabPanel.jsx
```diff
@@ -91,8 +97,9 @@ export default {
         <div
           v-show={this.active}
           aria-hidden={!this.active}
+          aria-labelledby={this.ariaLabelledby}
```
cowills, line 100: this should point to the id of a label element that describes this set of tabs, right? should we generating that label element here instead?

darintrimillos: The ariaLabelledby value here is used to generate the tab id as well. Is that what you mean?

src/components/tabs/CdrTabPanel.jsx
darintrimillos: I thought it was clearer to require the id, since we use it to link the tab panel back to the tab, but it didn't need to happen. I can add that back.

src/components/tabs/CdrTabPanel.jsx
cowills: ah yr right i just phenomenally mis-read this diff it seems :+1:

PR #800, opened 2020-03-11
button, cta, rating color updates

cowills:
## Description

Updating button, CTA, and everything that isn't an SVG in rating.

- icon-only button with border will be handled as a separate ticket
- need new SVG assets from design in order to update the rating stars

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

### A11y:
- [x] Meets WCAG AA standards

### Documentation:


src/components/button/styles/CdrButton.scss
```diff
@@ -1,26 +1,3 @@
-/* ==========================================================================
```
mhewson, line 1: generally I think the idea of a TOC is something to be preserved, this should be an artifact that helps ensure the completeness of the styles contained within. 
what reasons would we have to remove this?

mhewson (dismissed):

src/components/button/styles/CdrButton.scss
cowills: we don't keep them up to date. not all of our CSS files have TOCs. you can't actually navigate from the TOC directly to a style definition. our style files should be small anyways. 

happy to discuss it at dev critique if anyone feels strongly about keeping them

mhewson: Yea lets talk about what we could do in the future

mhewson (approved):

src/components/button/styles/CdrButton.scss
cowills: adding a card to dev critique :+1:

PR #806, opened 2020-03-16
fix(tabs): add aria-selected specificity

cowills:
`[aria-selected]` selector will still apply to `aria-selected="false"`, this makes it so disabled tabs don't get the active styling


darintrimillos (approved):
I noticed this right as you were opening the PR. Thank you.

PR #811, opened 2020-03-19
Link a11y unit tests

lancetharper:
## Description

- add skip link, aria-current, and active page indicator to kitchen sink nav
- Fixes errors with duplicate keys in examples
- Fixes unit test checking for content that wasn't doing anything before

I question whether we actually want to do what that test is saying. Do we want to just add 'Link Text' if nothing is in the slot? I don't even know if that case would ever actually happen, and if it did would that text even be helpful? 

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards


cowills:
does a link with no content even render anything to the page? i guess a screen reader would still detect something.... still seems like it would be better if it read the href out than just "Link text", which i suspect is what it would do by default to a link with no content?

lancetharper:
> does a link with no content even render anything to the page? i guess a screen reader would still detect something.... still seems like it would be better if it read the href out than just "Link text", which i suspect is what it would do by default to a link with no content?

What if the href is blank or an internal page link (so `#` or `#some-id`)? I'm just not sure that this is an a11y requirement that we need to guard against because while we check the box of having something...is it really better than the possibility of having nothing?

I also don't really see anyone using an empty link so it will probably never even come into play...

Thoughts @mhewson ?

cowills (dismissed):

cowills (approved):

PR #812, opened 2020-03-19
refactor(button): more obvious click action for testing

lancetharper:
I didn't see much for button a11y or unit tests that needed updated since we already check for providing button type, etc. This just adds a click counter where we just had a console.log before so it's easier to see


mhewson:
looks good - there will be some additional needs to update tokens or some future work for focus state contrast but other than that - looks good

mhewson (approved):

PR #822, opened 2020-03-30
Cdr 1670 - updates to modal

darintrimillos:
## Description
Modal updates, including: 
- spacing and styling updates
- fullscreen@671px

### Design:

### Cross-browser testing:
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE11
- [x] iOS
- [x] Android

~Added/updated backstop tests~

### Unit testing:

### A11y:
- [x] Meets WCAG AA standards

### Documentation:
~API docs created/updated~
~Examples created/updated~
Separate ticket for docs updates.


src/components/modal/CdrModal.jsx
```diff
@@ -249,7 +276,7 @@ export default {
             role="dialog"
             aria-modal={!!opened}
             aria-label={label}
-            {...{ attrs: this.dialogAttrs || {} }}
+            {...{ attrs: this.dialogAttrs }}
```
lancetharper, line 288: Do we need to be able to put attrs elsewhere? If not I'd rather just pass all attrs to this and get rid of the prop

src/components/modal/CdrModal.jsx
```diff
@@ -67,7 +71,16 @@ export default {
       };
     },
     dialogClass() {
-      return `${this.style['cdr-modal__dialog']} ${this.size}`;
+      return `${this.style['cdr-modal__dialog']}`;
+    },
+    scrollMaxHeight() {
+      const vertSpace = this.fullscreen ? 32 : 80;
```
lancetharper: Are these magic numbers (32 & 80) meaningful? We have tokens available in js if that makes sense to use here

src/components/modal/CdrModal.jsx
```diff
@@ -93,6 +109,15 @@ export default {
     document.removeEventListener('keydown', this.keyHandler);
   },
   methods: {
+    measureContent() {
+      this.$nextTick(() => {
+        this.totalHeight = window.innerHeight;
+        this.fullscreen = window.innerWidth < 672;
```
lancetharper: How was 672 decided since it isn't one of our normal breakpoints?

src/components/modal/CdrModal.jsx
```diff
@@ -279,9 +309,9 @@ export default {
                     <cdr-button
                       id="close-modal-button"
                       class={this.style['cdr-modal__close-button']}
-                      icon-only={true}
-                      on-click={onClick}
-                      aria-label="close"
+                      icon-only
+                      with-background={true}
```
cowills, line 322: missing the on-click handler here, close button isn't doing anything

src/components/modal/CdrModal.jsx
cowills: agreed, we should calculate these numbers based on tokens rather than hardcoding

src/components/modal/CdrModal.jsx
cowills: agreed, im guessing 672 is "some breakpoint - some spacing token" and we should calculate that rather than hardcoding it

src/components/modal/CdrModal.jsx
lancetharper: That's probably a breaking change though...so it's not really a big deal

src/components/modal/CdrModal.jsx
```diff
@@ -73,11 +74,19 @@ export default {
     dialogClass() {
       return `${this.style['cdr-modal__dialog']}`;
     },
+    verticalSpace() {
+      // contentWrap vertical padding
+      const fullscreen = (Number(CdrSpaceOneX.split('rem')[0]) * 10) * 2; // 32
```
cowills: the 2 here is still a magic number imo, is there no pure token value that would work here?

lancetharper: 10 is for converting from rem. Can js not use rem? I've never considered that. Do we need to make the js tokens px?

cowills:
@lancetharper @darintrimillos @mhewson  this should be ready to review.

We will definitely want to test this out in SSR, as i suspect we might need to make some of the computed props check for the presence of window first or have some fallback for non-browser envs. Seeing some warnings when i run the tests but not sure if it'll be impactful. But the easiest way to do that is to merge to next and cut an alpha

lancetharper (approved):
As long as things are updated in `mounted` it should work. Further testing to double check is definitely needed though

PR #833, opened 2020-04-09
Color a11y redux

cowills:
DONE:
- button
- link
- cta

FOLLOW UP PR:
- checkbox/radio/select/input

tested in all the browsers. need a doc update for the new link prop.


mhewson:
Looks like the icon is not inheriting the hover/focus text color updates in button

mhewson (approved):

PR #849, opened 2020-04-22
Really fix esm tokens

cowills:


mhewson (approved):

mhewson:
so what was the deal?

cowills:
> so what was the deal?

the usual esm vs. cjs nightmare. still not 100% sure if the problem is on our end or in vunit/febs but this solves it  🤷 seems to have to do with cdr-tokens.esm using named exports, whereas ESM packages that use default exports like lodash-es and clsx are fine.

mhewson:
it never ends

PR #856, opened 2020-05-13
feat(card): add configurable tag prop defaulting to `article`

cowills:
- should `article` be the default here? or would `div` be "safer"?


mhewson (commented):
thanks - I'll add back the docs

mhewson (approved):

PR #857, opened 2020-05-13
Drop ie11

cowills:
- deletes ie11
- updates browserslist config to match febs https://git.rei.com/projects/FEDPACK/repos/rei-front-end-build-configs/browse/shared/index.js#8-15
- deletes nonvariable font fallbacks (looks like our support matrix nicely lines up with variable font support 😹 https://caniuse.com/#feat=variable-fonts)

long term we should refactor our build to pull the browserslist from a single place (and also get the official REI browserslist package updated so we can use that and be sure we are in sync with febs)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -19,10 +19,9 @@ module.exports = function (api) {
       'Firefox > 64',
       'iOS > 11',
       'Safari >= 9',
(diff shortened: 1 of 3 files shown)


mhewson (approved):
Nice!

PR #862, opened 2020-06-02
fix(card): update card link underline logic

cowills:
currently failing on chromedriver which is fixed in a previous PR


mhewson (approved):
is this preferred over adding the standalone modifier to the link? This seems specific to the card title rather than any link that ends up being in a card - but that may be edge case thinking

cowills:
> is this preferred over adding the standalone modifier to the link? This seems specific to the card title rather than any link that ends up being in a card - but that may be edge case thinking

This class must be manually added to the link by the consumer when composing the card. IDK we could just update the doc site guidance for card to say "use `<cdr-link modifier="standalone" class="cdr-card__link">`" instead. 🤷

mhewson:
Thinking more about this it seems that this would be ok for now as standalone is really supposed to be semantically meaningful  - 

I still like the idea of a component or text utility for card-title that would handle this - but that can be v-sometime

PR #866, opened 2020-06-05
fix(breadcrumb): update breadcrumb to properly handle scopedSlot ref/tag

cowills:
CUS thread: https://rei.slack.com/archives/CA58YCGN4/p1591310986207800

`$refs` are only available in the scope in which they are rendered. so if you use the scopedSlots option with breadcrumb, then CdrBreadcrumb.jsx does not have access to `$refs.firstBreadcrumb` because that ref is rendered in the consuming component and not in breadcrumb itself. (i get why vue is like this...but, ugh)

In pagination we do this weird little dance to get access to refs rendered in scopedSlots: https://github.com/rei/rei-cedar/blob/next/src/components/pagination/CdrPagination.jsx#L334-L344 BUT for whatever reason similar logic does not seem to be working in breadcrumb to focus the first element.

Gotta use `li *` here since we have no idea what will be passed into the scopedSlot (i.e, a vs. vue-router-link vs. who knows what).

😪

### DIFF
--- a/src/components/breadcrumb/CdrBreadcrumb.jsx
+++ b/src/components/breadcrumb/CdrBreadcrumb.jsx
@@ -93,7 +93,6 @@ export default {
           /
         </span>) : '';
 
-        const ref = index === 0 ? 'firstBreadcrumb' : null;
         const isLink = index < this.items.length - 1;
         const LinkTag = isLink ? 'a' : 'strong';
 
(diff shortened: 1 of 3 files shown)


mhewson (approved):
we should push for prioritizing the removal of this ellipsis pattern

PR #868, opened 2020-06-05
Breadcrumb fixes

cowills:
- on actionable elements:
- - use link-hover token on hover 
- - use text-secondary on rest
- makes spacing between delimiters space-half-x
- replaces ellipses SVG icon with spaced periods: `. . .`

### DIFF
--- a/src/components/breadcrumb/__tests__/__snapshots__/CdrBreadcrumb.spec.js.snap
+++ b/src/components/breadcrumb/__tests__/__snapshots__/CdrBreadcrumb.spec.js.snap
@@ -18,17 +18,12 @@ exports[`CdrBreadcrumb renders correctly 1`] = `
         aria-label="show 1 more navigation level"
         class="cdr-breadcrumb__ellipses"
       >
-        <svg
+        <span
           aria-hidden="true"
-          class="cdr-icon cdr-breadcrumb__ellipses-icon"
-          role="presentation"
-          viewBox="0 0 24 24"
-          xmlns="http://www.w3.org/2000/svg"
+          class="cdr-breadcrumb__ellipses-icon"
         >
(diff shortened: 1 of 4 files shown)


src/components/breadcrumb/__tests__/__snapshots__/CdrBreadcrumb.spec.js.snap
```diff
@@ -18,17 +18,12 @@ exports[`CdrBreadcrumb renders correctly 1`] = `
...
-          role="presentation"
-          viewBox="0 0 24 24"
-          xmlns="http://www.w3.org/2000/svg"
+          class="cdr-breadcrumb__ellipses-icon"
         >
-          <path
-            d="M17.5 22a1.5 1.5 0 1 1 0-3 1.5 1.5 0 0 1 0 3zM12 22a1.5 1.5 0 1 1 0-3 1.5 1.5 0 0 1 0 3zm-5.5 0a1.5 1.5 0 1 1 0-3 1.5 1.5 0 0 1 0 3z"
-          />
-        </svg>
+          . . .
```
mhewson, line 25: may want to use … (&#8230; or &hellip;) horizontal ellipsis, used to indicate an omission or a pause (instead of ...)

mhewson (approved):

src/components/breadcrumb/__tests__/__snapshots__/CdrBreadcrumb.spec.js.snap
cowills: the design request was to have a space between each dot, so in that case we'd need a new SVG for that. The ellipsis element is aria-hidden and there is an aria-label on the container so not sure if it matters?

PR #874, opened 2020-06-05
Patterns - links

mhewson:
Link doc updates


docs_src/assets/css/src/docs.css
```diff
@@ -715,7 +715,7 @@ h1[id] {
   margin: 0;
   border: 0.1rem solid #eee;
   border-left-width: 0.5rem;
-  border-radius: 0.3rem;
+  border-radius: -0.3rem;
```
cowills, line 718: are these negative border radii correct?

docs_src/elements/typography.html
```diff
@@ -2,7 +2,8 @@
 layout: default
 title: Typography
 slug: typography
-lead: "An overview of Typography: headings/families, labels, and much more."
+lead: "Text container used for any text element such as paragraphs, headings, and lists. Establishes vertical spacing and
+optimizes the reading experience"
```
cowills, line 6: is this line-break intentional? not sure if it affects styling at all

docs_src/assets/css/src/docs.css
mhewson: yea the heading sizes changed and moved the icon alignment down with it

docs_src/elements/typography.html
mhewson: yea intended - no it does not effect the display

docs_src/assets/css/src/docs.css
cowills: but is a negative border-radius even something that works?
https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius seems to indicate that negative values are invalid

PR #882, opened 2020-06-17
update generator, refactor examples

cowills:
- removes "composition" option from component generator
- renames `examples` exports to match component name pluralization
- update dev router to pull examples from `examples` rather than importing in 2 places

Doing a lil cleanup while scaffolding the alert component 😀


src/components/examples.js
```diff
@@ -1,54 +1,51 @@
...
+  quote,
+  radio,
+  rating,
+  select,
+  table,
   tabs,
-  texts,
+  text,
   utilities,
 };
```
mhewson, line 51: ohh thanks!

mhewson (approved):

src/components/examples.js
mhewson: Always nice to find a sec to get to some of the fine tuning

PR #883, opened 2020-06-17
feat(alert): prototype alert component

cowills:
Screen Shot 2020-06-17 at 4 18 40 PM

SCSS structured so it can be dropped into the cedar 1 alert markup (AFAIK 🙈 ) not yet turned into component variables

some questions:

- What should the default text style be? (consumers could override via the content they pass in)
- Should the text color use the corresponding alert color token? (success/warning/info/error)
- What should the padding/spacing be?
- Should we center the icon vertically in the alert container? or keep it inline with the text? Any addition spacing needed between the icon and the text?
- Do we need responsive size modifiers?

### DIFF
--- a/src/components/alert/styles/CdrAlert.vars.scss
+++ b/src/components/alert/styles/CdrAlert.vars.scss
@@ -1,0 +1,37 @@
+@mixin cdr-alert-base-mixin() {
+  position: relative;
+  padding: $cdr-space-half-x;
+  @include cdr-text-default;
+}
+
+@mixin cdr-alert-info-mixin() {
+  background-color: $cdr-color-background-info;
+  border: 1px solid $cdr-color-border-info;
+  svg {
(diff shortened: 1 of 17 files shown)


src/components/alert/styles/CdrAlert.vars.scss
```diff
@@ -0,0 +1,37 @@
+@mixin cdr-alert-base-mixin() {
```
mhewson, line 1: One thing to consider on these, at least with the old float grid would be the z-index as many time these are overlaying other content

mhewson (approved):
Looks good! thanks for taking the auto icon out -

src/components/alert/styles/CdrAlert.vars.scss
cowills: I don't see any z-index applied to the alerts in patterns, do you have an example of that?

PR #885, opened 2020-06-22
feat(form-group): prototype form group component

cowills:
Screen Shot 2020-06-22 at 11 36 28 AM

simple wrapper that applies fieldset/legend container to form components


mhewson:
My thought is that we will want to turn off the border for the fieldset and allow users to set a cdr-text utility class on the legend

mhewson (approved):

cowills:
> My thought is that we will want to turn off the border for the fieldset and allow users to set a cdr-text utility class on the legend

In that case we'll need to just make the legend a slot so that users have full control over it. the text utility file is gigantic and we shouldn't bind non-cdr-text components to it

PR #886, opened 2020-06-25
Form group iteration

cowills:
designs: https://www.figma.com/file/wIdzSDusa2E21U3bhepjV2/Field-Sets?node-id=0%3A1

## formGroups
- adds label slot for override (and example to kitchen sink)
- sets default text styling for legend
- removes fieldset border
- removes default fieldset styling
- adds spacing between legend and content (consumers can override if needed by targeting `legend`)
- adds bottom margin (seems smarter to add this as a default and let consumers override by adding a class to their formGroup, than to require every consumer of formGroup to manually apply spacing)

## misc
- updates checkboxes dev example to use formGroup instead of fieldset
- sets default value for alert type

### DIFF
--- a/src/components/formGroup/styles/CdrFormGroup.vars.scss
+++ b/src/components/formGroup/styles/CdrFormGroup.vars.scss
@@ -1,4 +1,18 @@
 @mixin cdr-form-group-base-mixin() {
-  border: 1px solid $cdr-color-border-primary;
-  // TODO: styling for label text?
+  @include cdr-text-heading-sans-300;
+  border: none;
+
+  margin-inline-start: 0;
+  margin-inline-end: 0;
+  padding-block-start: 0;
+  padding-inline-start: 0;
+  padding-inline-end: 0;
+  padding-block-end: 0;
+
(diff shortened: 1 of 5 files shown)


src/components/formGroup/styles/CdrFormGroup.vars.scss
```diff
@@ -1,4 +1,18 @@
 @mixin cdr-form-group-base-mixin() {
-  border: 1px solid $cdr-color-border-primary;
-  // TODO: styling for label text?
+  @include cdr-text-heading-sans-300;
+  border: none;
+
+  margin-inline-start: 0;
```
mhewson, line 5: would there be performance value in adding the writing-mode value pair?

mhewson (approved):

src/components/formGroup/styles/CdrFormGroup.vars.scss
cowills: not sure i understand how thats relevant? do you have an example/link?

PR #891, opened 2020-07-07
Button cta merger

cowills:
brand == primary
light == secondary

- adds "dark" and "sale" color options to button
- adds elevated prop
- adds `icon-left` and `icon-right` slots. `icon-left` is functionally the same as the `icon` slot. Would deprecate the `icon` slot but we kind need it for `icon-only` buttons 🙃 
- removes `cdr-button__icon` utility class, instead targets `svg` and uses the `icon-` slot to apply correct styling to the icon
- adds `vertical-align: middle` to the button base class to ensure correct alignment. For some reason icon-left was causing the buttons to align to the top whereas icon-right was causing buttons to align to the bottom. Was mostly noticeable if an icon-left and icon-right button where placed next to each other. Since it's applied to the base element consumers could override if necessary.


mhewson:
How much of a need there is for us to support left and right icons in this component? It seems like it starts to stray a bit into composition land for a limited use-case. would it be just as easy for the user to add an icon and space?

cowills:
> How much of a need there is for us to support left and right icons in this component? It seems like it starts to stray a bit into composition land for a limited use-case. would it be just as easy for the user to add an icon and space?

to maintain how button currently works the consumer would need to add the appropriate padding-left/right to the button as well as the margin-right/left to the icon itself, which seems messier than just having the icon slots. I'm not sure why exactly we do it that way with changing the padding when an icon is present, I can play around with what it looks like to remove that and do the spacing manually.

src/components/button/styles/vars/CdrButton.vars.scss
```diff
@@ -225,12 +351,28 @@
   @include cdr-button-base-text-mixin-medium;
   padding: $cdr-space-inset-one-x-squish;
   width: auto;
+
+  &.cdr-button--has-icon-left {
```
cowills, line 356: @mhewson they'd also need to adjust the spacing for each different button size if we made it manual

mhewson:
yea not a deal breaker, it just seems that its less in line with our best practices - I'm also fine going with it the way it is and backlogging a future breaking change

mhewson (dismissed):

mhewson (approved):

PR #895, opened 2020-07-15
feat(rating): round rating display to nearest decimal, not quarter

cowills:
https://issues.rei.com/browse/CDR-1917

### DIFF
--- a/src/components/rating/__tests__/CdrRating.spec.js
+++ b/src/components/rating/__tests__/CdrRating.spec.js
@@ -7,6 +7,7 @@ describe('CdrRating', () => {
       propsData: {
         rating: 3.2323,
         count: 100,
+        href: 'rei.com'
       }
     });
     expect(wrapper.element).toMatchSnapshot();
@@ -22,6 +23,7 @@ describe('CdrRating', () => {
     expect(wrapper.vm.whole).toBe(3);
     expect(wrapper.vm.remainder).toBe('50');
     expect(wrapper.vm.rounded).toBe(3.5);
+    expect(wrapper.vm.displayRating).toBe(3.4);
     expect(wrapper.vm.empties).toBe(1);
 
     wrapper.setProps({rating: 3.122227});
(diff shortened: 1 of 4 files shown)


src/components/rating/examples/Ratings.vue
```diff
@@ -60,7 +74,7 @@
     />
     <div class="line-height-reset">
       <cdr-rating
-        rating="3.66"
+        rating="3.63"
```
mhewson, line 77: is there still a need for the rounded prop or should everything be moved to the displayRating - UGC is only interested in the 10's being displayed

src/components/rating/__tests__/CdrRating.spec.js
```diff
@@ -22,27 +23,31 @@ describe('CdrRating', () => {
     expect(wrapper.vm.whole).toBe(3);
```
mhewson, line 23: per the UGC convo today, they are interested in having the decimal place always present - 3.0 vs 3

src/components/rating/examples/Ratings.vue
cowills: yep we need that to figure out how many stars to show

src/components/rating/__tests__/CdrRating.spec.js
mhewson: and I'm not 100% sure but I believe the ask would be to move from .75 to .8 @NEWeber can you confirm?

cowills: already did the update from `.75` to `.8`. Just pushed a commit that makes `3` render as `3.0`

mhewson (approved):

src/components/rating/__tests__/CdrRating.spec.js
NEWeber: > and I'm not 100% sure but I believe the ask would be to move from .75 to .8 @NEWeber can you confirm?

Yes, that's right

PR #897, opened 2020-07-16
tabby scrolly grady

cowills:
Some preliminary work to help figure out how to build the scroll gradient component in a re-usable way

- re-maps some deprecated tokens
- updates tabs to use actual elements for gradient rather than before/after (so we can bind the background color to the gradient in the JSX)
- updates tabs to consider margin-left when calculating the width of the header elements
- updates tabs to allow for passing the background-color it is rendered on to ensure the scroll gradients render properly


mhewson:
yer on a roll today

mhewson (dismissed):

PR #900, opened 2020-07-21
A11y modal fix

cowills:
- removes role='presentation' from modal (https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html)
- changes role='presentation' to role='img' for icon (https://www.deque.com/blog/creating-accessible-svgs/)

### DIFF
--- a/src/components/modal/CdrModal.jsx
+++ b/src/components/modal/CdrModal.jsx
@@ -265,7 +265,6 @@ export default {
           },
         )}
         ref="wrapper"
-        role="presentation"
       >
         <div class={clsx(this.style['cdr-modal__outerWrap'], wrapperClass)}>
           <div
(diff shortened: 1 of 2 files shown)


src/components/alert/__tests__/__snapshots__/CdrAlert.spec.js.snap
```diff
@@ -7,7 +7,7 @@ exports[`CdrAlert matches snapshot 1`] = `
 >
   <svg
     class="cdr-icon"
-    role="presentation"
+    role="img"
```
mhewson: I'm wondering if we should add a role at all? sometimes the role presentation tag will be correct - when this graphic presents no reason to describe it as is the case with things like error alerts and such that already provide visible text. If we do want to move forward with role: image will be also be providing a means of exposing / modifying the title and description of the SVG's

mhewson (dismissed):

src/components/alert/__tests__/__snapshots__/CdrAlert.spec.js.snap
cowills: Sounds like we need more investigation before was can do this ticket, gonna revert this change for now

PR #901, opened 2020-07-22
Link button

cowills:
- Added `inset` and `size` props to link. Using `inset` by itself uses default medium size.
- Added responsive size options for button links. Sizing only triggers if `inset="true"` is also passed.
- Added button text styling to inset/size links


mhewson:
I think the inset class should also apply the default button text weight - great questions above

mhewson (approved):

PR #904, opened 2020-07-30
feat(popover): initial popover imported from fedpack

cowills:
This should be in a good state to merge this initial work in. Can always comment out the `export` if this isn't ready for summer release.

I think to implement the open/close animations, as well as fully support all the a11y reqs (ability to hover over an open tooltip, and setting aria-labelledby on the trigger pointing to an ID on the popover) will need to refactor this component a bit to make it so it's always rendered on the page but is simply hidden when it is not shown.

### DIFF
--- a/src/components/popover/CdrPopover.jsx
+++ b/src/components/popover/CdrPopover.jsx
@@ -50,7 +50,7 @@ export default {
   },
   computed: {
     positionClass() {
-      return this.style[`cdr-popover__${this.pos}`];
+      return this.style[`cdr-popover__wrapper--${this.pos}`];
     },
     cornerClass() {
       if (this.corner) return this.style[`cdr-popover__corner--${this.corner}`];
@@ -157,7 +157,12 @@ export default {
   },
   render() {
     return this.opened ? (
-      <div class={clsx(this.positionClass, this.cornerClass, this.triggerClass)}>
+      <div class={clsx(
+        this.style['cdr-popover__wrapper'],
+        this.positionClass,
+        this.cornerClass,
+        this.triggerClass)}
+      >
         <div
           class={this.style['cdr-popover']}
           role="dialog"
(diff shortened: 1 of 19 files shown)


src/components/popover/examples/Popover.vue
```diff
@@ -0,0 +1,180 @@
...
+        @click="togglePopover"
+        :icon-only="true"
+        aria-label="information"
+      >
+        <icon-information-fill />
+      </cdr-button>
+      <!-- TODO: export directive to handle hover/focus bindings? -->
+      <cdr-button
+        v-else
+        @mouseover="open = true"
```
cowills, line 115: having to set mouseover/mouseleave/focus/blur for a tooltip seems like a hassle for consumers. I think as part of the refactoring will want to handle this using a utility class, and have the CSS control the open/close logic on focus/hover. That should also make it easier to implement that "hovering over an active tooltip" feature

mhewson: yea I like that, it seems to make sense especially when we can introduce limited utilities that most users would want to use

src/components/popover/CdrPopover.jsx
```diff
@@ -0,0 +1,216 @@
...
+        }
+      });
+    },
+    addHandlers() {
+      this.keyHandler = this.handleKeyDown.bind(this);
+      document.addEventListener('keydown', this.keyHandler);
+      this.clickHandler = this.handleClick.bind(this);
+      document.addEventListener('click', this.clickHandler);
+    },
+    calculatePlacement() {
```
mhewson, line 106: I wonder if this is something we would want to add as a generic utility at some point?

mhewson (dismissed):

src/components/popover/CdrPopover.jsx
cowills: the click logic or the placement logic?

mhewson (dismissed):

PR #907, opened 2020-08-07
fix(card): fix image/border-radius clipping in card

cowills:
fixes the bug where the top left/right corners of a full width image inside a card would clip outside of the rounded border-radius.

overflow: hidden seems to make the most sense here, as a scrolling card would be very weird. But consumers can also override that attr if they really need that


src/components/card/styles/vars/CdrCard.vars.scss
```diff
@@ -7,6 +7,7 @@
   width: 100%;
   transition: box-shadow $cdr-duration-2-x $cdr-timing-function-ease;
   cursor: pointer;
+  overflow: hidden;
```
mhewson, line 10: rad

mhewson (approved):

PR #919, opened 2020-08-31
feat(grid): add responsive options for xs breakpoint

cowills:
- adds xs responsive options
- updates to new mq syntax

Could publish this as a minor version bump since it only adds functionality.

i think these media queries should technically be mq-only not mq-up, but that would be a big breaking change
`<cdr-row cols="6 12@sm">` currently turns into 6@xs, 12@sm/md/lg, whereas all of our other responsive options apply only at the specified breakpoints which would turn that into 6@xs/md/lg 12@sm. Might be worth assessing how many people are using the responsive aspects of grid and possibly making this consistent in a future release.

Like Grid, Image is also missing responsive props for XS and is using mq-up media queries, but it's a lil weird in that it applies the breakpoint classes via custom props and specifically calls out that it applies to mq-up: https://github.com/rei/rei-cedar/blob/next/src/components/image/CdrImg.jsx#L53-L79

### DIFF
--- a/src/components/grid/styles/CdrCol.scss
+++ b/src/components/grid/styles/CdrCol.scss
@@ -270,10 +270,206 @@ $colPct12: calc(12 / #{$rowColumns} * 100%);
   }
   */
 
+
+  /* COL EXTRA SMALL
+  ========================================================================== */
+
+  @include cdr-xs-mq-up {
+    &_span1\@xs {
+      flex-basis: $colPct1;
+      width: $colPct1;
+      max-width: $colPct1;
+    }
+
+    &_span2\@xs {
+      flex-basis: $colPct2;
+      width: $colPct2;
+      max-width: $colPct2;
+    }
+
+    &_span3\@xs {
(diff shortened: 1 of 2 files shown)


mhewson:
yea this all seems to need some work - grid could use some complexity removed

mhewson (approved):
I wonder if people would find it helpful to load each of the responsive breakpoints individually or prefer the all in one for this pattern?

cowills:
> I wonder if people would find it helpful to load each of the responsive breakpoints individually or prefer the all in one for this pattern?

Thats something Daniel has brought up, that we can dynamically load style sheets based on the breakpoint of the screen. Pretty sure that would need to get baked into febs/crampon to work though...would have to investigate more. 

We could break out the CSS files for responsive components into default/xs/sm/md/lg, and have the original component CSS file load all 5 of those files, that way consumers could have the option to optimize it and nothing existing would break.

mhewson:
yea that makes sense as a first pass

PR #926, opened 2020-09-23
switch min-width to max-width, allow overrides

cowills:
Im pretty sure the `min-width: 286px` bit was an artifact from the original adventures popover implementation. Was wiring up the tooltip to a forms POC and noticed it does this when the popup message is short

Screen Shot 2020-09-23 at 3 13 38 PM

Seems to make more sense for the popup to be as wide as it's content with a max-width.

max-width should also probably be changed to something in our space scale (256px would be `cdr-space-four-x` * 4)

Also updated popover/tooltip to accept a `contentClass` like modal does so they can override whatever default we set if needed.

### DIFF
--- a/src/components/popover/CdrPopover.jsx
+++ b/src/components/popover/CdrPopover.jsx
@@ -35,6 +35,10 @@ export default {
       type: String,
       required: true,
     },
+    contentClass: {
+      type: String,
+      required: false,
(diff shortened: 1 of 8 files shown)


mhewson:
I wonder if max-width should be something based on the viewport?

PR #928, opened 2020-10-05
tab focus/underline alignment

cowills:
- adds an invisible ::before element to the tab headers containing a bolded copy of the header text to prevent content from shifting when bolded
- switches li/a structure to buttons
- uses getBoundingClientRect for underline measurement, offsetLeft/offsetWidth are integers and can cause the underline to be off by a pixel or two
- increases the underline timeout in mounted to ensure it renders properly


mhewson (approved):
looks good!

PR #936, opened 2020-10-13
move type inheritance logic out of reset and onto italic/strong

cowills:
Since we only really need the font inheritance logic for italic/strong, just moves that logic into those classes. Backstop tests pass 100% when making this change.

This seems to cover our main use case for the cdr-text nesting/inheritance logic, which is someone highlighting a few words within a larger cdr-text element

Only 2 projects are using these modifiers so should have minimal impact if any.
https://git.rei.com/plugins/servlet/search?q=%22modifier%20strong%22
https://git.rei.com/plugins/servlet/search?q=%22modifier%20italic%22

### DIFF
--- a/src/css/reset.scss
+++ b/src/css/reset.scss
@@ -102,13 +102,4 @@ h5.cdr-text,
 h6.cdr-text {
   @include cdr-text-default;
   margin: 0;
-
-  .cdr-text {
-    font-weight: inherit;
-    font-size: inherit;
-    line-height: inherit;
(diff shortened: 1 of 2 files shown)


mhewson:
I think the problem was with things like dynamic data - something like expert advice where they may have the ability of setting the main font stack but then not be able to target specific <em>or <strong> tags - but if this works for that usecase also - so much the better

mhewson (approved):

PR #937, opened 2020-10-13
Grid gutter

cowills:
reverts this PR which added XS modifiers for grid: https://github.com/rei/rei-cedar/pull/919

- Since the grid modifiers are breakpoint-and-up, XS modifiers don't really make sense since thats functionally the same thing as passing an un-breakpoint-ified value i.e, `<cdr-row gutter="none@xs">` does the same thing as `<cdr-row gutter="none">`, both would set gutter to none at all breakpoints

- Another challenge is that the grid has responsive gutter behavior built into it. The "default" grid behavior is a 16px gutter @xs and @sm and a 32px gutter @md and @lg, so the breakpoint modifiers conflict with the default responsive behavior. There's currently no way to do `gutter="none@xs"` with the default behavior at the other 3 breakpoints.

Instead, I added "small" (16px) and "medium" (32px) gutter size options to go along with xxs and none. The "default" grid behavior is functionally `gutter="small@xs medium@md"`. That way someone can do gutter="none small@sm medium@md" to get the desired effect of "no gutter @xs, default behavior otherwise"

This seems like the best option for now, as it resolves the initial support request (https://rei.slack.com/archives/CA58YCGN4/p1598904816037900) without any breaking changes and with the smallest bundle size impact. Since it sounds like we want to simplify the grid in the future, probably makes sense to leave these components as they are and eventually deprecate them in favor of a simpler css-grid based solution. 

TODO:
- update grid docs API
- note how grid breakpoints work (or verify that we call that out already)

### DIFF
--- a/src/components/grid/CdrRow.jsx
+++ b/src/components/grid/CdrRow.jsx
@@ -44,15 +44,15 @@ export default {
       ),
     },
     /**
-     * Defines gutter size. Default gutter size is 16px @xs and @sm and 32px @md and @lg.
-     * Possible values: {none, xxs}.
+     * Defines gutter size. Default gutter size is 16px @xs and @sm (small) and 32px @md and @lg (medium).
+     * Possible values: {none, xxs, small, medium}.
      * Also accepts responsive values with `@breakpoint`: "none@md"
      */
     gutter: {
       type: String,
       validator: (value) => propValidator(
         value,
-        ['none', 'xxs'],
+        ['none', 'xxs', 'small', 'medium'],
       ),
     },
     /**
--- a/src/components/grid/examples/Grid.vue
+++ b/src/components/grid/examples/Grid.vue
@@ -580,6 +580,90 @@
           tag="h4"
           modifier="subheading"
         >
+          gutter="small"
+        </cdr-text>
+        <cdr-row
+          cols="3"
+          gutter="small"
+        >
+          <cdr-col>
+            <div class="row-demo">
+              3
+            </div>
+          </cdr-col>
+          <cdr-col>
+            <div class="row-demo">
+              3
+            </div>
+          </cdr-col>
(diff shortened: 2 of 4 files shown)


src/components/grid/CdrRow.jsx
```diff
@@ -44,15 +44,15 @@ export default {
...
+     * Defines gutter size. Default gutter size is 16px @xs and @sm (small) and 32px @md and @lg (medium).
+     * Possible values: {none, xxs, small, medium}.
      * Also accepts responsive values with `@breakpoint`: "none@md"
      */
     gutter: {
       type: String,
       validator: (value) => propValidator(
         value,
-        ['none', 'xxs'],
+        ['none', 'xxs', 'small', 'medium'],
```
mhewson, line 55: should these be sm, md or maybe update xxs to extra-small?

mhewson (dismissed):

src/components/grid/CdrRow.jsx
cowills: i think that gets confusing with the breakpoints. small@sm vs. sm@sm etc.
and i'd also rather not make a breaking change for no reason. plus this feature will probably be rarely used anyways.

mhewson: that makes sense however it still seems odd to have xxs, small, medium as a set - where are xs and large? why the shorthand for xxs if not for small and medium

cowills: xxs is 2px, small is 16px, medium is 32px.
i don't have any strong opinions on what to call them, i wouldn't have picked "xxs" in the first place because it looks like a breakpoint. we could just make it "none small medium large"

mhewson (dismissed):

mhewson (approved):

PR #939, opened 2020-10-19
Delete spring deprecated

cowills:
Removing a bunch of stuff that was deprecated >2 releases ago

- deletes CSS utils deprecated in spring
- deletes data-table
- removes deprecated compact modifier from radio/checkbox
- removes deprecated label prop from accordion
- removes deprecated tab event


mhewson (dismissed):
whoot

mhewson (approved):

PR #941, opened 2020-10-19
update font path

cowills:


danlope-rei (changes requested):
Files are at slightly different paths, see https://github.com/rei/rei-cedar/commit/e054ada17ad8c13673301fb23c1c3349f57687ce for reference

mhewson (dismissed):

PR #956, opened 2020-11-18
comp var cleanup, form design review

cowills:
- refactors component variables for all the form elements. Trying to remove as much positioning logic as i can from the comp vars so each mixin corresponds to a single element and is less impacted by markup structure. 
- adds component variables for radio/checkbox figure (still questionable how useful/practical that will be....)
- addresses design review feedback


mhewson (approved):
Nice!

PR #963, opened 2020-12-14
add github action for CI

cowills:
github actions gives you 2,000 free build minutes a month for open source projects. 
travis currently takes about 3 minutes to build cedar (less for tokens/docs), so we shouldn't ever hit that limit (unless we're pushing changes 30+ times a day every day for a month straight)

if this works here will add it to the other repos 🎉 

- runs build, lint, and tests on any PRs against next or master
- adds deploy script for kitchen sink/gh-pages


mhewson (approved):
fingers crossed!

PR #964, opened 2020-12-16
feat(grid): create css-grid prototype component

cowills:
Screen Shot 2020-12-16 at 11 32 30 AM

creates basic css-grid wrapper component so i can start working on doc site examples for parity with row/col.

https://css-tricks.com/snippets/css/complete-guide-grid/

### DIFF
--- a/src/components/examples.js
+++ b/src/components/examples.js
@@ -7,6 +7,7 @@ import card from 'componentsdir/card/examples/Cards';
 import checkbox from 'componentsdir/checkbox/examples/checkboxes';
 import cta from 'componentsdir/cta/examples/Cta';
 import formGroup from 'componentsdir/formGroup/examples/FormGroup';
+import gridTwo from 'componentsdir/gridTwo/examples/GridTwo';
 import grid from 'componentsdir/grid/examples/Grid';
 import icon from 'componentsdir/icon/examples/Icons';
 import image from 'componentsdir/image/examples/Images';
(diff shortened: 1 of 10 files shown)


src/components/examples.js
```diff
@@ -7,6 +7,7 @@ import card from 'componentsdir/card/examples/Cards';
 import checkbox from 'componentsdir/checkbox/examples/checkboxes';
 import cta from 'componentsdir/cta/examples/Cta';
 import formGroup from 'componentsdir/formGroup/examples/FormGroup';
+import gridTwo from 'componentsdir/gridTwo/examples/GridTwo';
```
mhewson, line 10: should we be more specific that this is cssGrid rather than flexgrid?

src/components/gridTwo/styles/vars/CdrGridTwo.vars.scss
```diff
@@ -0,0 +1,8 @@
+@mixin cdr-grid-two-base-mixin() {
+  display: grid;
```
mhewson, line 2: amazing that this replaces all that gunk of the old system

mhewson (dismissed):
This is looking great

src/components/examples.js
cowills: yeah i figure we can rename stuff later. this is about as complex as this component will get so it shouldn't be hard to change

PR #969, opened 2020-12-21
chore(text): force consumers to manually load cdr-text utility classes

cowills:
- deletes text utility import from CdrText. This forces consumers to load the `text.css` utility file if they want to use the text modifiers.
- moves cdr-text specific code out of reset into `cdr-text.css`. That way we still have the `1 JS component` : `1 css file` relationship.

This seems like a good step towards deprecating those utilities, as its much simpler to just say "stop using cedar utilities" than explaining the caveats about CdrText really using the text utility file and etc.


mhewson:
yea this seems logical - can we add explicate documentation on the text section of the doc site

PR #972, opened 2020-12-23
feat(modal): add dialog-class prop to modal

cowills:


src/components/modal/CdrModal.jsx
```diff
@@ -284,7 +281,6 @@ export default {
           <div
             ref="modal"
             class={clsx(this.style['cdr-modal__contentWrap'], dialogClass)}
-            id={modalId}
```
cowills, line 287: already gets bound as part of dialogAttrs...and also `modalId` does not exist

PR #986, opened 2021-02-16
Winter buggies

cowills:
- [x] fix modal safari ios bug (need to verify once this is on staging)
- [x] remove text fade from modal
- [x] breadcrumb a11y expanded bug
- [x] non-linked pagination

modal safari bug is propbably due to `transitionend` being weird in ios. i think with this fix the modal content display logic should no longer be tied to anything CSS animation related. Also deleted the overflow text fade stuff since we know it's a violation and anyone overflowing modals is doing something weird and un-user friendly anyways.

breadcrumb a11y bug seems to be a NVDA bug with focusing elements that just entered the page. will see if a timeout instead of a nextTick solve the issues. 

Seems like for pagination as long as we allow customizing the aria-label (to set it to be "Pagination for reviews section" or something in the UGC case) then we can render buttons instead of links and the rest of the markup should just work. Will need to double check that later.

### DIFF
--- a/src/components/breadcrumb/CdrBreadcrumb.jsx
+++ b/src/components/breadcrumb/CdrBreadcrumb.jsx
@@ -3,6 +3,8 @@ import CdrIcon from '../icon/CdrIcon';
 import modifier from '../../mixins/modifier';
 import style from './styles/CdrBreadcrumb.scss';
 
+// BREADCRUMB ISSUE!!!
+
 export default {
   name: 'CdrBreadcrumb',
   components: {
@@ -125,9 +127,9 @@ export default {
   methods: {
     handleEllipsisClick() {
       this.truncate = false;
-      this.$nextTick(() => {
+      setTimeout(() => {
         this.$el.querySelector('li *').focus();
-      });
+      }, 1);
     },
   },
   render() {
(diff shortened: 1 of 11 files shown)


src/components/breadcrumb/CdrBreadcrumb.jsx
```diff
@@ -125,9 +127,9 @@ export default {
   methods: {
     handleEllipsisClick() {
       this.truncate = false;
-      this.$nextTick(() => {
+      setTimeout(() => {
```
mhewson, line 130: what is the timeout doing? allowing for the motion?

mhewson (approved):

src/components/breadcrumb/CdrBreadcrumb.jsx
cowills: my theory is that the v-show on the truncated items hides them from NVDA and the nextTick runs too soon for it to properly announce them. Once this is on the kitchen sink i'll see if the team that reported the bug can verify.

we could also hide them some other way, having them be present in the DOM but visibly and aria-hidden till the ellipsis is clicked. but a 1ms timeout is more or less a nextTick so if this works :+1: from me.

mhewson: ah ok that makes sense - or we could just not hide them :p

PR #989, opened 2021-02-19
Accorddion unwrapped

cowills:
- remove focus ring from chippies
- add `unwrap` prop to accordion group which renders the accordions in an unwrapped state with no buttons/padding/etc.
- `unwrap` accepts a boolean or a list of breakpoints


cowills:
@mhewson should be good to review now

mhewson (approved):

PR #995, opened 2021-03-03
primary fill color of inputs is inaccessible on some backgrounds -

mhewson:
switching to secondary for the success, info, warning colors and error for the error bg

### Unit testing:
- [X] Sufficient unit test coverage (see unit test best practices for ideas)

### A11y:
- [X] Meets WCAG AA standards


cowills:
all the form components have a background prop, can add more options to it if needed (probably should make a mixin for it too). We can add additional background prop options as a minor version bump so no need to tie it to a quarterly release

PR #997, opened 2021-03-05
fix(input): fix issue with popover/tooltip in input info-action slot

cowills:
Screen Shot 2021-03-04 at 2 42 05 PM

apparently putting an element that does lots of `transform` weirdness like popover/tooltip into a slot that also uses `transform` to center it's content can cause issues 😆


mhewson:
🤦 the joy - good catch!

PR #1007, opened 2021-03-11
chore(lint): update lint config, add a11y plugin

cowills:
adds vue-a11y plugin. leaves it commented out as it currently has a false positive on form components.
the tree shaking plugin is useless, the other 2 plugins i used weren't being used. 
We don't use the `standard` JS syntax rules, and the only "promise" we use is vue.$nextTick.


mhewson (approved):
thanks for talking a look!

PR #1010, opened 2021-03-18
feat(number): better support for numeric input, default autocorrect off

cowills:
Enhances support for numeric inputs:

- for some reason `<input type="number">` neither creates a number-only input, nor does it auto-launch the "number" keyboard on mobile browsers. 
- instead we need to use `pattern` to restrict to numbers only, along with `novalidate` so that the pattern doesn't trigger native input validation
- we can use `inputmode` "numeric" to ensure the proper numeric keyboard is launched on mobile.

Adds default auto-correct attrs:
- Almost all of the form pattern examples specify that auto-correct/auto-capitalize should be disabled. I think it makes sense to default these attributes off and let consumers override them if needed. (`autocorrect`, `autocapitalize`, `spellcheck`)

Not sure if there are any a11y concerns here?

### DIFF
--- a/src/components/input/CdrInput.jsx
+++ b/src/components/input/CdrInput.jsx
@@ -87,6 +87,18 @@ export default {
     baseClass() {
       return 'cdr-input';
     },
+    inputAttrs() {
+      const isNum = this.type === 'number';
+      return {
+        pattern: isNum && '[0-9]*',
+        inputmode: isNum && 'numeric',
+        novalidate: isNum,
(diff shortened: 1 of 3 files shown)


mhewson:
I have been noticing this here - https://codesandbox.io/s/cedar-input-testing-rytl5?file=/App.vue - I have been trying to capture everything here - https://issues.rei.com/browse/CDR-2051

mhewson (approved):

PR #1011, opened 2021-03-19
fix(a11y): custom prop for aria-describedby to match native prop

mhewson:
Not huge, but confusing to have custom props deviate from the native name

### DIFF
--- a/src/components/modal/CdrModal.jsx
+++ b/src/components/modal/CdrModal.jsx
@@ -29,7 +29,7 @@ export default {
       required: false,
       default: true,
     },
-    ariaDescribedbBy: {
+    ariaDescribedby: {
       type: String,
       required: false,
       default: null,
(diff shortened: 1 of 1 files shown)


src/components/modal/CdrModal.jsx
```diff
@@ -29,7 +29,7 @@ export default {
       required: false,
       default: true,
     },
-    ariaDescribedbBy: {
```
mhewson, line 32: also a sadness typo

cowills: im pretty sure we can remove this prop entirely, unless we need special access to it there's no need to make props for native attrs. plus if the prop has been typo'd this whole time then... 😆

mhewson: 👍 yea probably should just go

src/components/modal/CdrModal.jsx
```diff
@@ -64,7 +59,6 @@ export default {
   computed: {
     dialogAttrs() {
       return {
-        'aria-describedby': this.ariaDescribedBy,
```
cowills, line 67: you still need to add this here, but you can pull the prop off attrs instead

$atttrs['aria-describedby'] should do it

cowills:
I will pull this into the safari modal bugfix work

PR #1015, opened 2021-04-06
WIP Safari modal fun time

cowills:
😪

### DIFF
--- a/src/components/modal/styles/CdrModal.scss
+++ b/src/components/modal/styles/CdrModal.scss
@@ -98,6 +98,28 @@ $modal-animation-duration: 150ms;
   }
 
   &__text-content {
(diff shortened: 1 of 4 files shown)


src/components/modal/styles/CdrModal.scss
```diff
@@ -98,6 +98,28 @@ $modal-animation-duration: 150ms;
   }
 
   &__text-content {
+    // might also be due to position fixed?
```
mhewson, line 104: 🎲 uggg

cowills: the best part is i have to spin up a try for every change i want to test since i cant repro this on the kitchen sink and browserstack doesn't work with the micro-site dev env 😆

PR #1017, opened 2021-04-07
fix naming of aria-describedby property

cowills:
this should make it so people can do `aria-describedby: 'foobar'` and have it work as expected


mhewson (approved):
oops forgot about this

PR #1020, opened 2021-04-12
Form error

mhewson:
## Description
Adding Role tag="status to the generic error message for input and props for the slot that allow for status or alert- Modal also updated to allow role being changed from default "dialog" to alertdialog  

### A11y:
- [x] Meets WCAG AA standards

### DIFF
--- a/src/components/input/CdrInput.jsx
+++ b/src/components/input/CdrInput.jsx
@@ -44,6 +44,11 @@ export default {
       type: String,
       required: true,
     },
+    errorRole: {
+      type: String,
+      required: false,
+      default: "status",
+    },
     /**
      * Removes the label element but sets the input `aria-label` to `label` text for a11y.
     */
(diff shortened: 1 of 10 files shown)


src/components/input/CdrInput.jsx
```diff
@@ -230,7 +235,7 @@ export default {
             </span>
         )}
         {this.error && (
-          <cdr-form-error error={this.error}>
+          <cdr-form-error role={this.errorRole} error={this.error}>
```
cowills, line 241: does this properly let you override the role in formError? can't recall how native attrs work if it's hardcoded in the child but overridden in the parent

mhewson: It did in the examples I added on input, I'm not sure if I caught all the use-cases though?

cowills:
select and formGroup also use formError so would need to be added in those places too presumably?

src/components/input/CdrInput.jsx
cowills: yeah i think thats how it works i just cant test it at the moment 😆  as long as you can pass `errorRole="foobar"` to an input and see foobar in the formError role then :+1:

mhewson:
Ah good call - I can take a look - the errors for now are defaulting to status for them - which they should be almost 99% of the time

cowills (approved):

PR #1030, opened 2021-05-24
fix(modal): add additional padding to modal content to accomodate scroll

cowills:
on modal open, we check to see if the content is overflowing AND what the size of the vertical scrollbar is. If overflowing and the vertical scrollbar has a width, that means a traditional permanent scrollbar is present. If overflowing and the vertical scrollbar has 0 width, that means a Mac floaty scrollbar will pop up on scroll.

If it's overflowing and there is a scrollbar present, add a small bit of padding. Otherwise the modal content juts up directly against the content.
If it's overflowing and there is no scrollbar present, add a bunch of padding so the floating scrollbar does not overlap with content.
if it's not overflowing add no padding

On Mac, you get the floaty scrollbar by default if using a trackpad. If you plug in a mouse you should see the permanent scrollbar. Though you need to re-open or resize the window after changing input device to get the padding to update to the new scroll type.
Browserstack won't seem to load an iOS device for me at the moment so can't test on mobile Mac yet. 
Screen Shot 2021-05-24 at 12 25 13 PM
Screen Shot 2021-05-24 at 12 24 58 PM

Here's what it looks like on the bopus modal (simulated by just adding the padding-right in on future):
Screen Shot 2021-05-24 at 12 31 55 PM
Screen Shot 2021-05-24 at 12 31 42 PM

### DIFF
--- a/src/components/modal/CdrModal.jsx
+++ b/src/components/modal/CdrModal.jsx
@@ -63,7 +63,6 @@ export default {
       offset: null,
       headerHeight: 0,
       totalHeight: 0,
-      scrollHeight: 0,
       isScrolling: false,
       hasScrollbar: false,
       fullscreen: false,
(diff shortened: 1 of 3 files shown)


mhewson (dismissed):
looks great

PR #1037, opened 2021-06-03
feat(input): dynamic aria-describedby, hidelabel sr-only

cowills:
- dynamically generates IDs for helper text slots in input/select and binds them to the aria-describedby prop on the form element
- still allows for consumer passing in aria-describedby
- updates `hideLabel` behavior to instead make the label element sr-only
- When an input, select, or formGroup is in an error state, automatically generates an `id` for the error element, sets `aria-invalid="true"` on the form component and `aria-errormessage` pointing to the generated error id

### DIFF
--- a/src/components/formGroup/CdrFormGroup.jsx
+++ b/src/components/formGroup/CdrFormGroup.jsx
@@ -10,6 +10,8 @@ export default {
     CdrFormError,
   },
   props: {
+
+    id: String,
     label: {
       type: String,
       default: '',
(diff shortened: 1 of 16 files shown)


cowills:
@mhewson can you take a look at this when you get a chance? i'm not sure if the error messaging should be linked to the input/select via aria-describe when the error is active, or if the `role="status"` on the error message is sufficient for screenreaders. Currently when the error is rendered it wipes out the "helper-text-bottom" slot, so it might need to have some logic like "link the error if its active, or link the helper-text-bottom if its present, or do nothing for that slot"

mhewson (dismissed):
Nice!

PR #1056, opened 2021-09-20
Cdr messaging

benjag:
Alpha branch with banner refactor and new toast component

### DIFF
--- a/src/components/banner/CdrBanner.jsx
+++ b/src/components/banner/CdrBanner.jsx
@@ -11,9 +11,9 @@ export default {
       type: String,
       validator: (value) => propValidator(
         value,
-        ['info', 'warning', 'success', 'error'],
+        ['info', 'warning', 'success', 'error', 'default'],
       ),
-      default: 'info',
+      default: 'default',
     },
   },
   data() {
@@ -31,7 +31,7 @@ export default {
     iconClass() {
       return this.modifyClassName(`${this.baseClass}__icon-left`, this.type);
     },
-    mainTheme() {
+    themeClass() {
       return this.modifyClassName(`${this.baseClass}__main`, this.type);
     },
     prominenceClass() {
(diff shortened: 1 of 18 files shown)


src/components/banner/examples/Banner.vue
```diff
@@ -1,27 +1,149 @@
 <template>
   <div>
     <h2>banner</h2>
+    <cdr-banner type="warning">
```
mhewson, line 4: I like the type prop - but I wonder if it is really a type and not a modifier... Applying type="button" or type="input" actually caries application consequence  whereas  type="error" will not even provide something like role="alert"

src/components/banner/styles/CdrBanner.scss
```diff
@@ -3,20 +3,120 @@
...
+    background-color: $cdr-color-background-message-default-02;
+  }
+}
+
+.cdr-banner__message {
+  grid-area: message;
+  padding: $cdr-space-half-x;
+}
+
+.cdr-banner__message-body {
```
mhewson, line 83: any reason not to `&-body` and so on here?

src/components/banner/styles/CdrBanner.scss
```diff
@@ -3,20 +3,120 @@
 
 .cdr-banner {
   @include cdr-banner-base-mixin;
+  display: grid;
```
mhewson, line 6: I'm a fan of alphabetization for our selector value pairs- open to other organizations though

src/components/banner/CdrBanner.jsx
```diff
@@ -25,14 +25,56 @@ export default {
     baseClass() {
       return 'cdr-banner';
     },
-    iconClass() {
+    borderClass() {
```
mhewson, line 28: do we need all these classes to be unique or can this be set via the parent modifier? 
something like .cdr-banner__info.cdr-banner--border

src/components/toast/styles/vars/CdrToast.vars.scss
```diff
@@ -0,0 +1,68 @@
...
+
+@mixin cdr-toast-base-mixin() {
+  position: relative;
+  border-radius: $cdr-radius-softer;
+  border-left: 4px solid;
+  box-shadow: $cdr-prominence-floating;
+  margin-bottom: $cdr-space-one-x;
+  @include cdr-text-default;
+  @include cdr-sm-mq-up {
+    width: 39rem;
```
mhewson, line 12: should this be a token?

mhewson (approved):

src/components/banner/styles/CdrBanner.scss
benjag: I'll clean this up

src/components/banner/examples/Banner.vue
benjag: Changing would introduce a breaking change to the existing banner API. I kept the same prop in toast for consistency

PR #1058, opened 2021-09-23
Cdr messaging updates

benjag:
### Toast
- Adds event handlers to prevent auto-dismissing on mouseover
- Toast now emits an open and close event
- Removed extraneous computed style props and refactored style asset so that sub elements inherit theme from parent.

### Banner
- Updates example page
- Removed extraneous computed style props and refactored style asset so that sub elements inherit theme from parent.

### DIFF
--- a/src/components/banner/CdrBanner.jsx
+++ b/src/components/banner/CdrBanner.jsx
@@ -25,14 +25,8 @@ export default {
     baseClass() {
       return 'cdr-banner';
     },
-    borderClass() {
+    typeClass() {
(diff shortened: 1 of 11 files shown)


src/components/banner/CdrBanner.jsx
```diff
@@ -25,15 +25,9 @@ export default {
     baseClass() {
       return 'cdr-banner';
     },
-    borderClass() {
+    typeClass() {
```
mhewson, line 28: nice

mhewson (approved):

PR #1062, opened 2021-09-29
Toast updates

benjag:
### Toast
- Fixed a bug where the toast would be auto-dismissed even when set to false after a mouseover event.
- Removes the border-radius style to match banner.


KenjiCrosland (approved):

mhewson:
would be nice to see the toast fired off with a chip - I think this will tend to be a common pattern as these tend to confirm a user setting and such

mhewson (approved):

PR #1063, opened 2021-10-08
Fix message icon

benjag:
Noticed that after my last refactor, the SVG fill style was a little too overzealous. This PR correctly scopes them to the icon slot.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@rei/cedar",
-  "version": "11.0.0-beta.2",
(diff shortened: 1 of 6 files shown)


mhewson:
is this fill something we can test out on the icon page itself? maybe adding this as in icon prop that banner is activating? 
My worry is that the slot leaves it open to any icon being used but we are not really looking at a large set of icons to see what the expected results will be - or I could just be missing that?

mhewson (approved):

PR #1069, opened 2021-10-15
Remove row col

benjag:
Removes CdrRow and CdrCol from the distributable


mhewson (approved):
see yea later col and row

KenjiCrosland (approved):

